PR #12, opened 2021-11-22
Use custom url for self-hosted

nathanaelhoun:
I suggest to replace the `ntfy.sh` url suggested in the code blocks to match the real url, to allow an easier copy-pasting if we selfhost.

This replacement has been tested with a custom port (first screenshot) and a classic port (80/443 for http/https) (second screenshot).

| On main page | On event page |
| :- | -: |
| image | image |

### DIFF
--- a/server/index.gohtml
+++ b/server/index.gohtml
@@ -67,19 +67,19 @@
         Here's an example showing how to publish a message using <tt>curl</tt> (via POST):
     </p>
     <code>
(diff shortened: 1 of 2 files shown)


nathanaelhoun:
Also, thanks for this "It just works" software, which looks good and works fine!

binwiederhier:
Awesome thank you.

PR #73, opened 2021-12-29
Update Dockerfile for new start command

ramonsnir:
https://ntfy-c92917546f9761e4.fly.dev/docs/deprecations/#running-server-via-ntfy-instead-of-ntfy-serve

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,4 +2,4 @@ FROM alpine
 MAINTAINER Philipp C. Heckel <philipp.heckel@gmail.com>
 
 COPY ntfy /usr/bin
-ENTRYPOINT ["ntfy"]
+ENTRYPOINT ["ntfy", "serve"]


binwiederhier:
@ramonsnir Do you know what the best practice is for a Docker that has multiple purposes. There are multiple commands now:

```
ntfy publish
ntfy subscribe
ntfy serve
```

And the docker image can be called just like that. 
I didn't know how people use this normally.

ramonsnir:
Hmm. I see what you're saying. I didn't realize some people are using the image as a CLI replacement.

Maybe instead of this PR (which I'm rejecting), the docs should describe how to run this in Docker non-interactively. For example, as of ~30 minutes ago, I'm running my self-hosted version on fly.io with a Dockerfile that looks something like this:
```
FROM binwiederhier/ntfy
COPY server.yml /etc/ntfy/server.yml
ENTRYPOINT ["ntfy", "serve"]
```

This allows me to just run the derivative Docker image without having to specify any command-line arguments to Docker.

binwiederhier:
I'd be (more than) happy to accept docs updates too. I'd actually love that :-D

PR #86, opened 2022-01-05
Adjust RPM scriptlets to work on RHEL-flavour OSes, too.

cmeis:
I suppose in prerm.sh the existing
```
[ "$1" = remove ]
```
should be a
```
[ "$1" = "remove" ]
```
too?

### DIFF
--- a/scripts/postinst.sh
+++ b/scripts/postinst.sh
@@ -6,7 +6,7 @@ set -e
 #
 # TODO: This is only tested on Debian.
 #
-if [ "$1" = "configure" ] || [ "$1" = "1" ]; then
+if [ "$1" = "configure" ] || [ "$1" -gt 1 ]; then
   if [ -d /run/systemd/system ]; then
     # Create ntfy user/group
     id ntfy >/dev/null 2>&1 || useradd --system --no-create-home ntfy
--- a/scripts/postinst.sh
+++ b/scripts/postinst.sh
@@ -6,7 +6,7 @@ set -e
 #
 # TODO: This is only tested on Debian.
 #
-if [ "$1" = "configure" ] || [ "$1" -gt 1 ]; then
+if [ "$1" = "configure" ] || [ "$1" -ge 1 ]; then
   if [ -d /run/systemd/system ]; then
     # Create ntfy user/group
     id ntfy >/dev/null 2>&1 || useradd --system --no-create-home ntfy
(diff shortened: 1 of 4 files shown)


scripts/postinst.sh
```diff
@@ -6,7 +6,7 @@ set -e
 #
 # TODO: This is only tested on Debian.
 #
-if [ "$1" = "configure" ] && [ -d /run/systemd/system ]; then
+if ( [ "$1" = "configure" ] || [ "$1" = "1" ] ) && [ -d /run/systemd/system ]; then
```
binwiederhier, line 9: Uhh does this work in a POSIX shell? I never know what works and what doesn't in /bin/sh

If you're sure then let's keep it. If not, maybe just do a nested if..

scripts/prerm.sh
```diff
@@ -2,7 +2,7 @@
 set -e
 
 # Stop systemd service
-if [ -d /run/systemd/system ] && [ "$1" = remove ]; then
+if [ -d /run/systemd/system ] && ( [ "$1" = remove ] || [ "$1" = "0" ] ); then
```
binwiederhier, line 5: Same as above with the parentheses.

As for the quotes I don't think it matters. But to keep it the same as the rest you can add quotes.

binwiederhier (approved):

cmeis:
As I too never know what works in a basic POSIX shell I switched the scriptlets to nested if statements just to be sure.

binwiederhier:
This looks great. 

Have you by any chance tested if the systemd service is restarted properly when you re-install the rpm and it's already running? I had a really bad time getting that to work on Debian. And since you're already working on it ...... you don't have to look into that though.

binwiederhier:
If not I'll merge this as is. Fantastic work. Thank you.

cmeis:
I suppose with re-install you mean an upgrade? This is the only case I can imagine where the service is running and it is re-installed. I simulated this by creating a separate RPM package with forged higher version numbers, so I had something to "test-upgrade".
On a first test, the service was NOT restarted after an upgrade (AlmaLinux 8 was used for testing).

I then read up on Fedora docs for this (https://docs.fedoraproject.org/en-US/packaging-guidelines/Scriptlets/) and adjusted the scripts. Subsequent tests showed that the service was correctly restarted on an upgrade.

What is to be seen (but not that important I think): During a downgrade the service is NOT restarted. I couldn't find a notice as to how to detect a downgrade in the scriptlet, but didn't give this search much time.

Hope this helps?

binwiederhier:
Awesome. Thanks for also fixing the restart thing!!

PR #148, opened 2022-02-17
Add authentification support for subscribing with CLI

### LINKED ISSUE #147: Add support for authentification with the CLI client
Currently, it's impossible to use the CLI client to subscribe to a topic that require authentication.

I use a self hosted private instance which require authentication, the CLI client can't subscribe to any topic on my instance without authentication support.

Authentication support for subscribing should be added to the CLI client and documentation updated in consequence.

lrabane:
I added `-u username[:password]` parameter to `ntfy subscribe` just like the one in `ntfy publish`.
I also added `user` and `password` key for each subscription in `client.yml`.

Close #147

### DIFF
--- a/client/config.go
+++ b/client/config.go
@@ -14,9 +14,11 @@ const (
 type Config struct {
 	DefaultHost string `yaml:"default-host"`
 	Subscribe   []struct {
-		Topic   string            `yaml:"topic"`
-		Command string            `yaml:"command"`
-		If      map[string]string `yaml:"if"`
+		Topic    string            `yaml:"topic"`
+		User     string            `yaml:"user"`
+		Password string            `yaml:"password"`
+		Command  string            `yaml:"command"`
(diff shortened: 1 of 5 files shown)


client/config.go
```diff
@@ -14,9 +14,11 @@ const (
...
 	Subscribe   []struct {
-		Topic   string            `yaml:"topic"`
-		Command string            `yaml:"command"`
-		If      map[string]string `yaml:"if"`
+		Topic    string            `yaml:"topic"`
+		User     string            `yaml:"user"`
+		Password string            `yaml:"password"`
+		Command  string            `yaml:"command"`
+		If       map[string]string `yaml:"if"`
 	} `yaml:"subscribe"`
```
binwiederhier, line 22: Oh cool, per subscription passwords. I like it.

binwiederhier (approved):

PR #152, opened 2022-02-24
Add support for iOS app

Copephobia:
Initial support for the iOS app.

TODOs:
- [ ] APNS payload - add sound/vibrate
- [ ] Send "poll_request"APNS payload
- [ ] Look into message/payload length limits (note: if payload is too large, APNS will not deliver the notification)

### DIFF
--- a/server/server_firebase.go
+++ b/server/server_firebase.go
@@ -13,6 +13,7 @@ import (
 
 const (
 	fcmMessageLimit = 4000
+	fcmApnsBodyMessageLimit = 100
 )
 
 // maybeTruncateFCMMessage performs best-effort truncation of FCM messages.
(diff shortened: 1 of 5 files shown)


server/server_firebase.go
```diff
@@ -88,6 +98,17 @@ func toFirebaseMessage(m *message, auther auth.Auther) (*messaging.Message, erro
 				data["attachment_expires"] = fmt.Sprintf("%d", m.Attachment.Expires)
 				data["attachment_url"] = m.Attachment.URL
 			}
+			apnsConfig = &messaging.APNSConfig{
+				Payload: &messaging.APNSPayload{
+					Aps: &messaging.Aps{
+						MutableContent: true,
+						Alert: &messaging.ApsAlert{
+							Title: m.Title,
+							Body:  maybeTruncateAPNSBodyMessage(m.Message),
```
Copephobia, line 107: The "body" of the push notification can contain the entire message, which would count doubly for the overall length of the APNS payload. I set a limit of 100 characters before truncating the notification "body" with ellipsis.

The message would not be changed (unless truncated for being too long).

Note: if the payload is too large (>4KB), APNS will simply reject / discard the notification, meaning it will never arrive on the iOS device.

binwiederhier:
The iOS app development sadly has been abandoned for now.

PR #173, opened 2022-03-17
Add examples for *arr notification scripts

nickexyz:
Something like this maybe?
Feel free to change, remove, add or do anything else.
If you rather want to include the scripts in your examples, that's fine too.

My first Github PR, hopefully it's correct. :)


binwiederhier:
Looks great. Thank you.

PR #178, opened 2022-03-19
Add Node-RED pictures and change ntfy URL to ntfy.sh

nickexyz:
Hopefully the pictures are in the right place.


binwiederhier:
I had forgotten about this PR. Thanks again!

PR #184, opened 2022-03-22
Go 1.18

binwiederhier:


binwiederhier:
This is on hold until staticcheck and IntelliJ catch up with their Go 1.18 support.

(See https://github.com/dominikh/go-tools/issues/1166)

PR #249, opened 2022-05-12
Add disable option to web-root

### LINKED ISSUE #238: Deny WebUI access to anonymous users
While the current ACL implementation allows you to deny subscribing and publishing for anonymous users, there's no way to prevent them from using the WebUI to connect to other servers. For a truly private server, anonymous users should be prevented from using the server at all

I have an idea on how this could be implemented:
1. Introduce a new ACL permission that controls WebUI access
2. Introduce a login screen at WebUI load
    - If WebUI access for anonymous users is allowed, the login screen is hidden
    - An added benefit is that you can use this authentication to forego asking users for credentials when adding local subscriptions later on

Curid:
Closes #238


server/config.go
```diff
@@ -126,5 +127,6 @@ func NewConfig() *Config {
 		VisitorEmailLimitBurst:               DefaultVisitorEmailLimitBurst,
 		VisitorEmailLimitReplenish:           DefaultVisitorEmailLimitReplenish,
 		BehindProxy:                          false,
+		DisableWebUI:                         false,
```
binwiederhier, line 130: Can you name this `WebEnabled` instead and make it `true` by default? That will invert the `if` statements.

cmd/serve_linux.go
```diff
@@ -55,6 +56,7 @@ var flagsServe = []cli.Flag{
 	altsrc.NewIntFlag(&cli.IntFlag{Name: "visitor-email-limit-burst", EnvVars: []string{"NTFY_VISITOR_EMAIL_LIMIT_BURST"}, Value: server.DefaultVisitorEmailLimitBurst, Usage: "initial limit of e-mails per visitor"}),
 	altsrc.NewDurationFlag(&cli.DurationFlag{Name: "visitor-email-limit-replenish", EnvVars: []string{"NTFY_VISITOR_EMAIL_LIMIT_REPLENISH"}, Value: server.DefaultVisitorEmailLimitReplenish, Usage: "interval at which burst limit is replenished (one per x)"}),
 	altsrc.NewBoolFlag(&cli.BoolFlag{Name: "behind-proxy", Aliases: []string{"P"}, EnvVars: []string{"NTFY_BEHIND_PROXY"}, Value: false, Usage: "if set, use X-Forwarded-For header to determine visitor IP address (for rate limiting)"}),
+	altsrc.NewBoolFlag(&cli.BoolFlag{Name: "disable-webui", Aliases: []string{"d"}, EnvVars: []string{"NTFY_DISABLE_WEBUI"}, Value: false, Usage: "if set, disables WebUI"}),
```
binwiederhier, line 59: Can you call this flag `disable-web`?

binwiederhier (approved):
LGTM

1. Can you add docs? Mostly the table, I don't think we need a special section for this in the config docs; just the table(s) are enough.
2. And the `server.yml` entry
3. And maybe a quick test for it.

Thanks

cmd/serve_linux.go
Curid: I didn't realize the flag didn't need to have the same name as the variable.

binwiederhier: Actually. Darn. This is tricky. Because this is the server.yml file config option. I wanted to name it `web-enabled`, but boolean flags with a true default don't work. I don't know how to solve this.

Curid:
Added docs and tests, my editor uses [gofumpt](https://github.com/mvdan/gofumpt) btw.

binwiederhier:
You're gonna hate me for this, but what about this: why not make the `web-root` option have three possible states:

- `app` -- / app root
- `home` -- / is landing page, /app is app root
- `disabled` -- web app and home are entirely disabled (NEW)

You could turn the `WebRootIsApp` config param into `WebRoot`.

My reasoning is this: I hate having new parameters. And I also hate having a parameter called "disable-x". While that's okay in the CLI, it's super weird in yml.

What do you think?

binwiederhier:
Looks great. Thank you for your contribution!

tokenwizard:
Hey @Curid, fancy finding you on this project, too. 
Am I missing something here? I just want to disable the webui and just use CLI/API and the Mobile Apps. 
I have my server.yml like this but I can still access the webui anonymously. 

Running via Docker method. 

Here is my docker-compose.yml:
```
version: "2.3"

services:
  ntfy:
    image: binwiederhier/ntfy
    container_name: ntfy
    command:
      - serve
    environment:
      - TZ=America/New_York    # optional: set desired timezone
    volumes:
      - ./var/cache/ntfy:/var/cache/ntfy
      - ./etc/ntfy:/etc/ntfy
      - ./server.yml:/etc/ntfy/server.yml
      - ./var/lib/ntfy:/var/lib/ntfy
    ports:
      - 80:80
    healthcheck: # optional: remember to adapt the host:port to your environment
        test: ["CMD-SHELL", "wget -q --tries=1 http://localhost:80/v1/health -O - | grep -Eo '\"healthy\"\\s*:\\s*true' || exit 1"]
        interval: 60s
        timeout: 10s
        retries: 3
        start_period: 40s
    restart: unless-stopped
```


Here is the web-root section of my server.yml:
image

I could be wrong, but it seems I can anonymously access the webui and even the Settings. I am even able to delete my "james" user anonymously, with no authentication required to do so. 
image

I would expect only an Admin to have that level of access.

tokenwizard:
I think I see now. With the option set to "disable" the root url does throw a 404 error. Adding /app top the end still loads the webui. But the Users don't show in the webui if I load it on another device, so it only shows my "james" user because that is stored locally int he web browser. 

Sound about right?

wunter8:
If adding /app at the end of the URL still loads the web app when it's set to disable, that is probably a bug.

But yes, you are correct about the local storage. The web app is just a simple client (like the Android app): you can open it and look around, but you cannot publish/subscribe to topics unless you put in auth credentials

Issue #112, opened 2022-01-20
Desktop client (Windows/Mac/Linux)


12nick12:
Hi, I would love a Windows client with push support

thefeiter:
This should be doable with python. Maybe not as a final solution, but afaik there is a library for python to send notifications on Windows.

binwiederhier:
Is this client supposed to send notifications or receive them? If it's the former, i can probably just compile the CLI for Windows.

Can you be a little more specific?

12nick12:
receive notifications and send them via push on the Windows client.

SMAW:
Maybe a tutorial like this can help?
Or using electron at all

https://jojozhuang.github.io/tutorial/converting-web-app-to-desktop-app-with-electron/

c33s:
i would love to subscribe to topics via a windows client. shouldn't it be possible to compile go for windows?

binwiederhier:
There are two things: 

* The CLI (excluding the server) can likely be easily compiled for Windows and Mac, then you can subscribe like described here https://ntfy.sh/docs/subscribe/cli/
* The web app (like this one: https://ntfy.sh/app) can be packaged into a desktop app as well, using electron or others. I played with this today and it's absolutely doable. 

Q: Which one would take precedent for you guys?

binwiederhier:
Note to self, also maybe look at https://github.com/tauri-apps/tauri, thanks `@Laurenz` (Matrix).

binwiederhier:
(from this afternoon, WIP which does not work: https://github.com/binwiederhier/ntfy/pull/239)

arminus:
Regarding the above question: What I'm looking for is an unobtrusive/low footprint systray thingy which tells me about new messages (I frequently forget to keep an ntfy browser tab open, yeah I know, can be pinned...).

Low footprint is kind of an issue with Electron, also not sure how well it would integrate with the systray and windows notification "requirement", so what I did yesterday was to prototype a small WPF app. I'm neither a .NET/WPF nor a C# expert so I'm not sure how far I will get with reasonable effort and energy but here goes a first proof of concept:

![NtfyWindowsClient](https://user-images.githubusercontent.com/683680/167290752-4be6bf10-b3dc-4812-a16f-492fef07f7c6.png)

[...]

PR #258, opened 2022-05-16
Use relative paths in WebUI

### LINKED ISSUE #256: Unable to host on a subdirectory.
Despite the base url option, when hosting somewhere that isn't the root directory, it will still load resources as if it was in the root directory.

Curid:
Closes #256

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -8,7 +8,6 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
-	"heckel.io/ntfy/log"
 	"io"
 	"net"
 	"net/http"
@@ -22,6 +21,8 @@ import (
 	"sync"
 	"time"
 	"unicode/utf8"
+
+	"heckel.io/ntfy/log"
 
 	"github.com/emersion/go-smtp"
 	"github.com/gorilla/websocket"
@@ -75,6 +76,9 @@ var (
 
 	//go:embed "example.html"
 	exampleSource string
+
+	//go:embed site/app.html
+	appHTML string
 
 	//go:embed site
 	webFs        embed.FS
@@ -317,10 +321,10 @@ func (s *Server) handleInternal(w http.ResponseWriter, r *http.Request, v *visit
 
 func (s *Server) handleHome(w http.ResponseWriter, r *http.Request, v *visitor) error {
 	if s.config.WebRootIsApp {
-		r.URL.Path = webAppIndex
-	} else {
-		r.URL.Path = webHomeIndex
+		return s.handleRoot(w, r, v)
 	}
+
+	r.URL.Path = webHomeIndex
 	return s.handleStatic(w, r, v)
 }
 
@@ -332,8 +336,7 @@ func (s *Server) handleTopic(w http.ResponseWriter, r *http.Request, v *visitor)
 		_, err := io.WriteString(w, `{"unifiedpush":{"version":1}}`+"\n")
 		return err
 	}
-	r.URL.Path = webAppIndex
-	return s.handleStatic(w, r, v)
+	return s.handleRoot(w, r, v)
 }
 
 func (s *Server) handleEmpty(_ http.ResponseWriter, _ *http.Request, _ *visitor) error {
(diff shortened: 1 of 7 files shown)


Curid:
It's a bit buggy at the moment. I tried to mark it as a draft.

binwiederhier:
Cool thank you! The `root = window.location.pathname` is a genius idea!

Few things:

1. Is the `./` really necessary? So can't I just do `<a href="app">` instead of `<a href="./app">`?
2. You committed `react-scripts`, which should not be in there. Can you remove that?
3. Be sure to try attachments too, and sounds and the `web-root = home` setting. I vaguely remember having issues with that.

Curid:
> Is the ./ really necessary? So can't I just do `<a href="app">` instead of `<a href="./app">?`

wtf, now it works without it.

> You committed react-scripts, which should not be in there. Can you remove that?

I made a few changes that don't seem necessary either.

```
+++ webpack.config.js
@@ -218,9 +218,7 @@ module.exports = function (webpackEnv) {
       pathinfo: isEnvDevelopment,
       // There will be one main bundle, and one file per asynchronous chunk.
       // In development, it does not produce real files.
-      filename: isEnvProduction
-        ? 'static/js/[name].[contenthash:8].js'
-        : isEnvDevelopment && 'static/js/bundle.js',
+      filename: 'static/js/bundle.js',
       // There are also additional JS chunk files if you use code splitting.
       chunkFilename: isEnvProduction
         ? 'static/js/[name].[contenthash:8].chunk.js'
@@ -609,7 +607,7 @@ module.exports = function (webpackEnv) {
         Object.assign(
           {},
           {
-            inject: true,
+            inject: false,
             template: paths.appHtml,
           },
           isEnvProduction
```

> Be sure to try attachments too, and sounds and the web-root = home setting. I vaguely remember having issues with that.

I still have to figure out why some static assets don't work.

`window.location.pathname` doesn't seem to work reliably if you refresh the page. We may have to add `#` to the root path and set root to that. `localhost/#/settings`

binwiederhier:
> window.location.pathname doesn't seem to work reliably if you refresh the page. We may have to add # to the root path and set root to that. localhost/#/settings. Or we could just remove the routing?

I don't want hash routing, and I definitely want to keep the routing. If you can't get it to work, then I'd much rather not support relative paths. I think that's why I gave up after an hour -- the routing....

binwiederhier:
The test file thing sounds horrible :-D I'd much rather prefer a different way. This has to be a solved problem, right?

Curid:
The root must be identifiable in some way, `/#/` or `/:/` If we only had a single sub directory depth then we could use the last `/`

binwiederhier:
https://ntfy.sh/mytopic must be the topic's address. What if you write something when the server initially loads?! Idk, just brainstorming.

Curid:
I came up with another to do it. It's a bit dirty but should be reliable.

Curid:
@binwiederhier I got everything working now. It works surprisingly well.

binwiederhier:
Just acknowledging that I saw this. I will likely not get to this for a while.

binwiederhier:
I'm just going to say it like it is, don't be mad, but the design and implementation of this is not right. I think I'll eventually need to make the whole thing work on a subdirectory for the Electron app (#239, #112), but I really don't want to probe or sleep or anything. The app should know where it's served from. This has to be a solved problem. 

How's this: When you request /subdir1/mytopic, the server returns the HTML of the topic page contains `const rootPath = "/subdir1"` somewhere in the rendered HTML. How the server figures out the root path I'm not quite sure. Maybe the proxy server sends `X-Root-Path` or something, idk. 

My point is: you should not guess and you should not sleep on the client side.

Curid:
> I'm just going to say it like it is, don't be mad

lol, don't worry I know it's bad. 

> Maybe the proxy server sends X-Root-Path

What if the path is injected into `index.html`?

Curid:
How about this?

amcraig:
Sorry for jumping on an old pull request, but I wanted to chime in saying subpath networking in the future would be tremendously helpful.

If I have enough time I might also be able to contribute a bit.

Congrats on the kid!

Curid:
Well, this is 15 hours I'll never get back.

PR #262, opened 2022-05-17
Docs/config: fix typo in private server example

MayeulC:
The `/etc/ntfy/server.yml` example was invalid yaml previously.

That's the config I'm using, so it works :)


binwiederhier:
Thank you

MayeulC:
Well, thank *you* for ntfy and UnifiedPush support :heart:

PR #270, opened 2022-05-26
fix(docker-security): use scratch container

### LINKED ISSUE #213: Use Dockerfile from scratch
You deploy an golang application, it is not necessary to use a complete alpine in it.

https://chemidy.medium.com/create-the-smallest-and-secured-golang-docker-image-based-on-scratch-4752223b7324

It makes the image more secure and smaller.

genofire:
fix #213

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM alpine
+FROM scratch
 MAINTAINER Philipp C. Heckel <philipp.heckel@gmail.com>
 
 COPY ntfy /usr/bin


genofire:
I believe, yes.
If you poke around please start another container with an image like alpine.

A container should not contain unneeded code - that is against the idea of container and a unneeded attack surface.

Alpine is a distro with many dependencies, they always as vulnerability. It is just time they become public.
If you do not release a new container with same ntfy version but every new alpine version. We have running an old alpine (with known vulnerabilities) with your latest ntfy version.
So we could get an security issue, just because you keep old base images in your container (e.g. current you used 3.15.4 and not 3.16.0).
So please drop this unneeded base-image.

binwiederhier:
Your pull request does not create a runnable Docker image for me, and I do not think this is a worthwhile effort. I'm sorry but I'd rather not spend more time on this. You are of course free to create your own ntfy image from the binary.

Thank you for your contribution.

```
$ docker run -it binwiederhier/ntfy:v1.23.0-amd64
docker: Error response from daemon: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: exec: "ntfy": executable file not found in $PATH: unknown.
```

genofire:
Your problem should be fixed.
Another problem could be your `CGO_ENABLED=1` for sqlite

Hypnotist1148:
This should be reopened again.

> Another thing would be to build the Images using the [CMD instruction](https://docs.docker.com/reference/dockerfile/#cmd), this would allow us to ommit the "serve" command.
> You can still use a different command by simply starting the container as you would now.

I will create a seperate issue for this request.

PR #282, opened 2022-05-29
Update the example docs with a Home Assistant notify example

poblabs:
I spent some time trying to find a config pattern that worked with Home Assistant. Finally found one and wanted to include it in the example docs.


binwiederhier:
Awesome. Thank you for your contribution. I'll merge it right away. Looks great.

PR #304, opened 2022-06-01
Docs docker

ksurl:
add info for non-root user and an example docker-compose.yml file in root


binwiederhier:
Looks great. Thank you.

PR #307, opened 2022-06-02
add timezone for docker install

ksurl:
add support for timezone

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,5 +3,7 @@ MAINTAINER Philipp C. Heckel <philipp.heckel@gmail.com>
 
 COPY ntfy /usr/bin
 
+RUN apk add --no-cache tzdata
+
 EXPOSE 80/tcp
 ENTRYPOINT ["ntfy"]
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -9,6 +9,7 @@ services:
     volumes:
       - /var/cache/ntfy:/var/cache/ntfy
       - /etc/ntfy:/etc/ntfy
+      - /etc/localtime:/etc/localtime:ro
     ports:
       - 80:80
     restart: unless-stopped
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -5,11 +5,12 @@ services:
     container_name: ntfy
     command:
       - serve
-    user: UID:GID # optional. Set custom user/group or uid/gid
+    environment:
+      - TZ=UTC    # optional: Change to your desired timezone
+    user: UID:GID # optional: Set custom user/group or uid/gid
     volumes:
       - /var/cache/ntfy:/var/cache/ntfy
       - /etc/ntfy:/etc/ntfy
-      - /etc/localtime:/etc/localtime:ro
     ports:
       - 80:80
     restart: unless-stopped
--- a/docs/install.md
+++ b/docs/install.md
@@ -239,11 +239,11 @@ docker run \
     serve
 ```
 
-With other config options, host-matched timezone, and non-root user (configured via `/etc/ntfy/server.yml`, see [configuration](config.md) for details):
+With other config options, timezone, and non-root user (configured via `/etc/ntfy/server.yml`, see [configuration](config.md) for details):
 ```bash
 docker run \
   -v /etc/ntfy:/etc/ntfy \
-  -v /etc/localtime:/etc/localtime:ro \
+  -e TZ=UTC \
   -p 80:80 \
   -u UID:GID \
   -it \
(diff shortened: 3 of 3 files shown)


binwiederhier:
I reverted this change because it fails the build pipeline for ARM. My assumption is that the alpine:latest image is an amd64 image and it cannot run `apk`. I think there is a way to solve this, but for this hotfix release I didn't have time.

Failed pipeline: https://github.com/binwiederhier/ntfy/runs/6718645921?check_suite_focus=true

ksurl:
hmm that's odd. I am using buildx for multi arch builds and my image also has tzdata install with no issues but I am not using goreleaser. just direct docker buildx action.

binwiederhier:
Goreleaser can do buildx apparently. https://goreleaser.com/customization/docker/

And you can set the platform flag too i think.

binwiederhier:
You may be able to do something like this: https://github.com/zinclabs/zinc/blob/main/Dockerfile#L6

egooner:
Any progress on this - using Delay send with absolute time not simple in Docker as it is using UTC and I am not in UTC!! Delay with offset is fine but would be nice to be able to use absolute time also.

kruton:
I was really confused because the docs were not reverted.

Ishrathh:
Time zone can be added using docker-compose, let me know if you would like me to share @egooner @kruton

egooner:
I have the following already set ... but to be honest cannot recall whether this did the trick or not but if I issue date in the console I get this so I think not!!
___
Fri Sep  8 18:46:27 UTC 2023
___
environment:
      - "TZ=Europe/London"
volumes:
      - "/etc/localtime:/etc/localtime:ro"

Ishrathh:
You can check for the correct time zone and there are 2 issues, both the machine running the container as well as the container have to be on the correct time zone. I believe you can manage it for your machine but for the docker instance here are my instructions:

#  docker-compose.yml 
```
version: "2.3"

services:
  ntfy:
    build: . # This will build dockerfile in current dir
    container_name: ntfy
    command:
      - /usr/bin/ntfy
      - serve
    volumes:
      - /var/cache/ntfy:/var/cache/ntfy
      - /etc/ntfy:/etc/ntfy
      - /etc/letsencrypt:/etc/letsencrypt
    ports:
      - 8080:80
    healthcheck: # optional: remember to adapt the host:port to your environment
        test: ["CMD-SHELL", "wget -q --no-check-certificate --tries=1 https://localhost:80/v1/health -O - | grep -Eo '\"healthy\"\\s*:\\s*true' || exit 1"]
        interval: 60s
        timeout: 10s
        retries: 3
        start_period: 40s
    restart: unless-stopped
```

#  Dockerfile
```
# Use the original ntfy image as the base
FROM binwiederhier/ntfy

# Install tzdata and tini packages
RUN apk update && apk add tzdata tini

# Set the timezone
ENV TZ=Asia/Colombo

# Use tini as the entrypoint
ENTRYPOINT ["/sbin/tini", "--"]
```

This is what I used, change the time zone to yours as needed. I also had to use tini which reaps zombie processes that were accumulated by 'ssl_client' process of ntfy. Also note that you can set the time zone in docker-compose as an environment variable OR in the Dockerfile. Dockerfile is easier if you don't plan on changing the time zone often

To check if the time zone of the container is correct you can do the following:
```
# sudo docker exec -it ntfy /bin/sh
# date
```

egooner:
Thanks for that @Ishrathh - I did your docker build example with the exception of the tini which did not seem to work for me!! 
Would be good to see Timezone built into the system rather than having to do this!! @binwiederhier  - any chance of incorporating the tzdata package into the build and supporting passing the TZ in the Environment variable. The solution above works but creates more effort on behalf of the users to do this.

Ishrathh:
Your welcome @egooner, if you would like me to assist you with the setup of tzdata without the tini part let me know :D

> I reverted this change because it fails the build pipeline for ARM. My assumption is that the alpine:latest image is an amd64 image and it cannot run `apk`. I think there is a way to solve this, but for this hotfix release I didn't have time.
> 
> Failed pipeline: https://github.com/binwiederhier/ntfy/runs/6718645921?check_suite_focus=true

@binwiederhier Would be awesome if we can get the time zone released on arch's that support it

binwiederhier:
How are we supposed to do that? Different docker files just because of this? That's seems little ridiculous.

egooner:
Not sure how your build process works @binwiederhier  (not very good on docker build etc) but I spotted this on conditional docker build not sure if it helps. [Docker File Conditions](https://stackoverflow.com/questions/43654656/dockerfile-if-else-condition-with-external-arguments)

binwiederhier:
Re-added for linux/amd64 in https://github.com/binwiederhier/ntfy/commit/8adb9ee633c5cca607892b9ab42b2685fa7fec73, see #894

Issue #319, opened 2022-06-11
Consider Including a Matrix Gateway endpoint as part of ntfy


MayeulC:
Hello, this is kind of awkward to ask as the use-case is quite specific, but at the same time it should be fairly self-contained.

### Rationale

The core issue is using UnifiedPush in conjunction with a Matrix Client and Server. The Matrix Client will pass a push URL to the Matrix server. However, the Matrix spec mandates some endpoints for push servers: https://spec.matrix.org/v1.2/push-gateway-api/

Basically, the spec [mandates](https://spec.matrix.org/v1.2/push-gateway-api/#post_matrixpushv1notify) the use of the ` /_matrix/push/v1/notify` path.

Therefore, one has to use an intermediate gateway before forwarding requests to ntfy. When users do not use their own, clients use default ones, which led to rate-limiting: #144 

#### Benefit to users

Currently, users can host a gateway themselves through the common-proxies project, or some other [rewrite proxy](https://unifiedpush.org/developers/gateway/#nginx).

That's one more step for users, who might feel discouraged from doing so.
Including that gateway would increase privacy, for users of the public ntfy instance, but even more so for users who host their own homeserver and ntfy instance, as push messages do not need to travel through third-party gateways.

### Needed changes

The set of features to support is relatively simple:

* Answer
  ```HTTP
   GET /_matrix/push/v1/notify
  ```
  With
  ```HTTP
  HTTP/1.1 200 OK

  {"unifiedpush":{"gateway":"matrix"}}
  ```
* Handle Pushes on that endpoint. From the spec:
  ```HTTP
  POST /_matrix/push/v1/notify
  
  {
  "notification": {
    "content": {
      "body": "I'm floating in a most peculiar way.",
      "msgtype": "m.text"
    },
    "counts": {
      "missed_calls": 1,
      "unread": 2
    },
    "devices": [
      {
        "app_id": "org.matrix.matrixConsole.ios",
        "data": {},
        "pushkey": "https://ntfy.sh/exampleTopicID?up=1",
        "pushkey_ts": 12345678,
        "tweaks": {
          "sound": "bing"
        }
      }
    ],
    "event_id": "$3957tyerfgewrf384",
    "prio": "high",
    "room_alias": "#exampleroom:matrix.org",
    "room_id": "!slw48wfj34rtnrf:example.com",
    "room_name": "Mission Control",
    "sender": "@exampleuser:matrix.org",
    "sender_display_name": "Major Tom",
[...]

PR #326, opened 2022-06-14
Matrix gateway

binwiederhier:


karmanyaahm:
It looks awesome to me! (Though I think https://ntfy.sh/docs/publish/#matrix-gateway is yet to be defined, as you said in https://github.com/binwiederhier/ntfy/issues/319#issuecomment-1157110685)

PR #327, opened 2022-06-15
Add user now supports reading password from an env var.

Kenix3:


binwiederhier:
@Kenix3 bump. you mentioned you wanted to switch this to `NTFY_PASSWORD=... ntfy user add phil`.

cmd/user.go
```diff
@@ -137,6 +139,7 @@ Examples:
 func execUserAdd(c *cli.Context) error {
 	username := c.Args().Get(0)
 	role := auth.Role(c.String("role"))
+	password := c.String("user")
```
binwiederhier, line 142: This should probably be "password"

binwiederhier:
Thank you

PR #341, opened 2022-06-22
allow custom intent in broadcast action without JSON

wunter8:
I noticed you could only change the intent of a broadcast action using the JSON publishing format. This allows you to specify the intent as a key/value pair in the Actions header

### DIFF
--- a/server/actions.go
+++ b/server/actions.go
@@ -186,6 +186,8 @@ func populateAction(newAction *action, section int, key, value string) error {
 			newAction.Method = value
 		case "body":
 			newAction.Body = value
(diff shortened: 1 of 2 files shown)


binwiederhier:
Good catch. Awesome!

To make it the perfect PR, can you

- Run `make fmt` to format the code
- Add an entry in the `releases.md` (including a thank you to youself :smile:)

wunter8:
Oops. Sorry about the formatting. My machine defaults to spaces instead of tabs, so it looked the same on my end, but I see the difference. I'll remember that for the future!

PR #343, opened 2022-06-24
all md files translated to spanish

johnfelipe:
Signed-off-by: felipe <ingenierofelipeurrego@gmail.com>


binwiederhier:
So at first I thought "oh wow that's a lot of effort", but then I looked and it seems that you (or the machine you used) even translated file names and things like "Go" or "package.json", which is silly and pretty much proves that this wasn't done by a human.

As a result, I'm closing this ticket.

johnfelipe:
You need show me Wich parts is bad translated and I will do it

Issue #69, opened 2021-12-29
Encrypted messages for Android app + CLI


binwiederhier:

binwiederhier:
I looked at this a little today, and sadly had to rule out `gpg`, because the library for Go is deprecated, and I don't want to build on deprecated technology. I also looked at `age` a while ago and even opened an issue here (https://github.com/FiloSottile/age/discussions/404) to ask for support in other languages; and sadly there is not enough support. 

I also played with `openssl` to see if I can encrypt and protect the integrity (~ AEAD-style), but I wasn't successful. 

I will try and play some more, but I believe I'll have to implement my own format. Something similar to what I have done here (https://syncany.readthedocs.io/en/latest/security.html#encrypting-new-files), though much simpler hopefully.

binwiederhier:
Here's the basic message structure: https://github.com/binwiederhier/ntfy/pull/354. This implements the exact same thing that Pushbullet does (https://docs.pushbullet.com/#encryption) and is even compatible with it. This was super easy. The trickier part is figuring out the UX of this. I'll have to think about that.

binwiederhier:
This can encrypt a message with a key, but a key is not a password. The password we'll still have to derive using pbkdf2. 

The question is now:
- Does each topic have its own password?
- Or is it somehow tied to the user? If so, pub-sub doesn't really work anymore, because the ciphertext is now different for each user.
- Where do we enter a topic password? What happens if it's wrong? 
Relatively easy would be:
- Each subscription has a password.
- The password is stored/entered in the subscription settings and stored locally
- If an encrypted message comes in, we derive the key using the password and use it to decrypt the message. If it succeeds, it's displayed. If not, we show (encrypted message)
- If an unencrypted message comes in, we do what we do now.

goalieca:
Password:

- minimum 6 characters, max 255.
- simple byte-string (implementations need to be consistent though, presumably utf-8 source, no null termination)
- can also be thought of as a generic secret (eg: could use yubikey to generate a strong key)

KDF:
- Should specify the URI is normalized. Non-normalized forms will derive different keys.
- I'm still thinking about specifying the number of rounds as a parameter? crypto agility would say yes. but then that would need to be stored alongside the password or in the jwe header along with kid. That's messy. That's the fun with 'dir' mode.

goalieca:
[...]

PR #354, opened 2022-07-01
WIP: Crypto stuff

binwiederhier:


examples/publish-php/publish-encrypted.php
```diff
@@ -0,0 +1,46 @@
+<?php
+
+$message = [
+    "message" => "Secret!",
+    "priority" => 5
+];
+$plaintext = json_encode($message);
+$key = deriveKey("secr3t password", "https://ntfy.sh/mysecret");
```
goalieca, line 8: i live the idea of having the fully qualified name here too as this will make the key context specific to the service and resource! That's great, as we always want unique derivations per context. But if i'm pedantic again, this will derive a different salt depending how the resource is represented (port :443 is set, trailing /, hostname aliases, missing http(s), ..)

goalieca (commented):
Pretty cool to see such easy implementations in these languages! That's gonna make things easy for devs.

nogweii:
I'd recommend looking at [libNaCl](https://nacl.cr.yp.to/) for implementing the cryptography: It's an awesome library that provides high level concepts and leaves dealing with the cryptographic primitives to the very smart people who designed the library. It's been ported to pretty much every language I can think of:

* Go: https://pkg.go.dev/golang.org/x/crypto/nacl
* Python: https://github.com/pyca/pynacl/
* JavaScript: https://www.npmjs.com/package/js-nacl
* PHP: https://www.php.net/manual/en/book.sodium.php (built-in as of PHP 7.2.0, for https://github.com/jedisct1/libsodium-php/ older PHPs)
* Ruby: https://github.com/RubyCrypto/rbnacl

binwiederhier:
Closed for now, will be reviewed and reused when #69 is finally being addressed

PR #356, opened 2022-07-03
Unix Socket Mode

koro666:
An attempt at adding an option to control the file mode of the Unix socket.


server/config.go
```diff
@@ -1,12 +1,14 @@
...
 
 import (
+	"io/fs"
 	"time"
 )
 
 // Defines default config settings (excluding limits, see below)
 const (
 	DefaultListenHTTP                           = ":80"
+	DefaultListenUnixMode                       = 0777
```
binwiederhier, line 11: Why would 777 be the default? Can't you use the 700 or just the default umask?

server/server.go
```diff
@@ -207,6 +207,13 @@ func (s *Server) Run() error {
 				errChan <- err
 				return
 			}
+			err = os.Chmod(s.config.ListenUnix, s.config.ListenUnixMode)
+			if err != nil {
+				s.unixListener.Close();
+				s.unixListener = nil
+				errChan <- err
+				return
+			}
```
binwiederhier, line 216: - Inline the call in the "if"
- no semicolon
- nilling is not necessary, since the program exits

``` go
if err := os.Chmod(s.config.ListenUnix, s.config.ListenUnixMode); err != nil {
  s.unixListener.Close()
  errChan <- err
  return
}
```

binwiederhier (changes requested):
This is fine, because it's small. Next time before you make changes, try to discuss it with me first, or even open a ticket so I can comment on it.

server/config.go
koro666: So that everything can connect to it by default, but I can change it.
As for the umask, I think it's already using it, which is why it defaults to 0755 without the `Chmod` call.
But for this particular case, the umask should be ignored, in favor of an explicit mode being set.
I could just set the service's umask to 0000 and "fix" my problem, but then every other file created by it would be world-writable.

koro666:
> This is fine, because it's small. Next time before you make changes, try to discuss it with me first, or even open a ticket so I can comment on it.

Noted.

I have made the changes in rebase, also changed the flag from `UintFlag` to `IntFlag` because it was not being read from `server.yml`.
Let me know if I should change the default mode as well.

PR #372, opened 2022-07-31
Client: default-user, default-password, default-command

### LINKED ISSUE #360: Default (user|password|command) in client.yml
Could it be possible to add the following top-level options to `client.yml`?

- `default-user`: the default value for `user` on a subscribed topic if none specified;
- `default-password`: the default value for `password` on a subscribed topic if none specified;
- `default-command`: the default value for `command` on a subscribed topic if none specified.

This would allow to save some copy-pasting in the event that a lot of subscribed topics use the same values.

Thanks.

wunter8:
fixes #360

### DIFF
--- a/client/client.yml
+++ b/client/client.yml
@@ -5,6 +5,12 @@
 #
 # default-host: https://ntfy.sh
 
+# Defaults below will be used when a topic does not have its own settings
+#
+# default-user:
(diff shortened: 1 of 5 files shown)


binwiederhier:
Awesome, thank you!

PR #373, opened 2022-08-01
Move `COMMIT` into a variable so it could be overridden if desired

cyqsimon:
This is so that when the source directory is not a git directory (as is the case with building from a tarball), it becomes possible to manually set the commit hash during build like so:
```sh
make VERSION=0.69.420 COMMIT=1337deadbeef cli-linux-server
```
---

Currently, as there is no way to override the commit manually, it's not possible to build a working binary from a tarball using `make`.

If you try anyway (trying is the first step to failure), since `git rev-parse --short HEAD` will fail silently, the binary will be built with `commit = ""`, which results in a runtime panic:
```
panic: runtime error: slice bounds out of range [:7] with length 0
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,6 @@
 MAKEFLAGS := --jobs=1
 VERSION := $(shell git describe --tag)
+COMMIT := $(shell git rev-parse --short HEAD)
 
 .PHONY:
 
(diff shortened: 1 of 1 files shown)


binwiederhier:
Thanks

PR #391, opened 2022-08-31
Add UptimeRobot example

youphyun:
This is my first pull request after a long time, so I hope I did everything right. I added an UptimeRobot example with 2 screenshots for ntfy. I hope it can be of use.


binwiederhier:
Thank you!

PR #393, opened 2022-08-31
unify unix-specific code

la-ninpre:
current code was tied to linux only, but it should work for any unix-like OS (e.g. BSDs).

i tested it on openbsd and can confirm that it works. on other OSes more testing is needed.

### DIFF
--- a/cmd/publish_linux.go
+++ b/cmd/publish_linux.go
@@ -1,8 +1,0 @@
-package cmd
-
-import "syscall"
-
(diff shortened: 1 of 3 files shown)


la-ninpre:
tested on netbsd, freebsd and dragonfly, everything works. only makefile requires some tweaking, but that's for another pr i guess.

binwiederhier:
Thanks!

PR #429, opened 2022-10-05
Allow Exempting IP Ranges

### LINKED ISSUE #423: Allow IP CIDRs in `visitor-request-limit-exempt-hosts`
For example, `visitor-request-limit-exempt-hosts=matrix.me.org,1.1.1.1,172.0.18.0/16`. Then, all of 172.0.18.0/16 will be exempt in addition to matrix.me.org and 1.1.1.1.

This will help with https://github.com/spantaleev/matrix-docker-ansible-deploy/pull/2135, where there is no way to exempt requests from a whole network.
Is this feature fine with you? If so, I can make a PR within the next few weeks.

karmanyaahm:
- Refactor visitor IPs and allow exempting IP Ranges

  Use netip.Addr instead of storing addresses as strings. This requires
  conversions at the database level and in tests, but is more memory
  efficient and efficient to compare; mainly it facilitates the following.
  
  Parse rate limit exemptions as netip.Prefix. This allows storing IP
  ranges in the exemption list. Regular IP addresses (entered explicitly
  or resolved from hostnames) are IPV4/32, denoting a range of just that one
  address.

~~**draft** still need to add tests and fix bugs.~~

However, is the change from string IPs to netip.Addr and netip.Prefix acceptable?

Closes #423

### DIFF
--- a/cmd/serve.go
+++ b/cmd/serve.go
@@ -306,31 +306,30 @@ func sigHandlerConfigReload(config string) {
 }
 
 func parseIPHostPrefix(host string) (prefixes []netip.Prefix, err error) {
-        //try parsing as prefix
-        prefix, err := netip.ParsePrefix(host)
-        if err == nil {
-                prefixes = append(prefixes, prefix.Masked()) // masked and canonical for easy of debugging, shouldn't matter
-                return prefixes, nil                         // success
-        }
+	//try parsing as prefix
+	prefix, err := netip.ParsePrefix(host)
+	if err == nil {
+		prefixes = append(prefixes, prefix.Masked()) // Masked returns the prefix in its canonical form,  the same for every ip in the range. This exists for ease of debugging. For example, 10.1.2.3/16 is 10.1.0.0/16.
+		return prefixes, nil                         // success
+	}
 
-        // not a prefix, parse as host or IP
-        // LookupHost forwards through if it's an IP
-        ips, err := net.LookupHost(host)
-        if err == nil {
-                for _, i := range ips {
-                        ip, err := netip.ParseAddr(i)
-                        if err == nil {
-                                prefix, err := ip.Prefix(ip.BitLen())
-                                if err != nil {
-                                        return prefixes, errors.New(fmt.Sprint("ip", ip, " successfully parsed as IP but unable to turn into prefix. THIS SHOULD NEVER HAPPEN. err:", err.Error()))
-                                }
-                                prefixes = append(prefixes, prefix.Masked()) //also masked canonical ip
-                        }
-                }
-        }
-        return
+	// not a prefix, parse as host or IP
+	// LookupHost forwards through if it's an IP
+	ips, err := net.LookupHost(host)
+	if err == nil {
+		for _, i := range ips {
+			ip, err := netip.ParseAddr(i)
+			if err == nil {
+				prefix, err := ip.Prefix(ip.BitLen())
+				if err != nil {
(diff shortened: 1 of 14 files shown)


binwiederhier:
This approach is perfectly fine. I'll leave some comments.

util/util.go
```diff
@@ -45,6 +46,16 @@ func Contains[T comparable](haystack []T, needle T) bool {
 	return false
 }
 
+// ContainsContains returns true if any element of haystack .Contains(needle).
+func ContainsContains[T interface{ Contains(U) bool }, U any](haystack []T, needle U) bool {
```
binwiederhier, line 50: 1. This is wild. I didn't know Go generics can do that.
2. This is oddly named
3. This is incomprehensible and very against my KISS rules (keep it simple)

Can you express this function in an easier way. It doesn't have to be generic if it is only used once, maybe? Maybe just `ContainsIP`?

binwiederhier (changes requested):

util/util.go
karmanyaahm: Changed to non-generic ContainsIP

server/server.go
```diff
@@ -642,8 +643,8 @@ func (s *Server) parsePublishParams(r *http.Request, v *visitor, m *message) (ca
 			return false, false, "", false, errHTTPBadRequestDelayTooLarge
 		}
 		m.Time = delay.Unix()
-		m.Sender = v.ip // Important for rate limiting
 	}
+	m.Sender = v.ip // Important for rate limiting
```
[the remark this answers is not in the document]
karmanyaahm: This was because of https://github.com/binwiederhier/ntfy/pull/429#discussion_r988526622. I thought there was a different bug, I'll make that not must and make this not record the IP for every message.

cmd/serve.go
```diff
@@ -303,6 +305,33 @@ func sigHandlerConfigReload(config string) {
 	}
 }
 
+func parseIPHostPrefix(host string) (prefixes []netip.Prefix, err error) {
+        //try parsing as prefix
+        prefix, err := netip.ParsePrefix(host)
+        if err == nil {
+                prefixes = append(prefixes, prefix.Masked()) // masked and canonical for easy of debugging, shouldn't matter
+                return prefixes, nil                         // success
+        }
```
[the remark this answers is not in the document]
karmanyaahm: For example, if someone enters 10.1.2.3/8, the canonical address is 10.0.0.0/8. It represents every address from 10.0.0.1 to 10.255.255.255. If the address is ever logged in the future, this would help the output be more consistent.

Should I explain this more in the comment?

binwiederhier:
Looks great. What's missing (since it's still in draft)?

cmd/serve_test.go
```diff
@@ -70,6 +72,22 @@ func TestCLI_Serve_WebSocket(t *testing.T) {
...
+		"201:be93::4a6/21": "201:b800::/21",
+	}
+	for q, expectedAnswer := range cases {
+		ips, err := parseIPHostPrefix(q)
+		require.Nil(t, err)
+		assert.Equal(t, 1, len(ips))
+		assert.Equal(t, expectedAnswer, ips[0].String())
+	}
+}
+
```
karmanyaahm, line 90: Should I just omit testing hostnames (because I'd have to mock the DNS resolver and all that)? The code is not all that complicated.

karmanyaahm:
server/server.go is already very thoroughly tested. I don't think there's anything meaningful I can add for this specific change.

PR #430, opened 2022-10-07
Delete server.yml

JayCraig2217:


binwiederhier:
I just closed this because I assumed that this was a spam PR, but I wanted to at least comment and ask if there is a reason behind this.

PR #434, opened 2022-10-09
allow empty password in client.yml

### LINKED ISSUE #374: [Bug] User with empty password in `client.yml` does not work
## Description

Using `ntfy` CLI, when trying to subscribe to a topic as a user with an empty password, it seems impossible to do so from the config file `client.yml`. Things I tried all result in `403 forbidden` authentication failures.

## Steps to reproduce

1. Set up a topic `empty-password-test` on the server with policy `deny`.
```sh
ntfy access everyone empty-password-test deny
```
2. Add a user `alice` with empty password and read access to `empty-password-test`.
```sh
ntfy user add alice # provide empty password
ntfy access alice empty-password-test ro
```
3. Run `curl -u "alice:" -s http://my.domain.tld/empty-password-test/json`, and observe that stream is opened successfully.
  - This verifies that the server has been set up correctly.
4. Set this up in `client.yml`.
```yml
subscribe:
    - topic: "http://my.domain.tld/empty-password-test"
      command: 'notify-send "$m"'
      user: "alice"
      password: ""
```
5. Run `ntfy sub -C`, and observe that authentication fails.
6. Run `ntfy sub -u "alice:" -C`, and observe that with the username&password override, authentication no longer fails.

## Miscellaneous

- I tried `password: ~`, `password: null`, `password: false`, and omitting the `password` field altogether. None worked, although `password: false` produced `401 unauthorized` instead of `403 forbidden`.
- Versions:
  - Linux x86_64 5.18.14, ntfy 1.27.2@69d6cdd, go 1.18.4

wunter8:
fixes #374

### DIFF
--- a/client/config.go
+++ b/client/config.go
@@ -12,14 +12,14 @@ const (
 
 // Config is the config struct for a Client
 type Config struct {
-	DefaultHost     string `yaml:"default-host"`
-	DefaultUser     string `yaml:"default-user"`
-	DefaultPassword string `yaml:"default-password"`
-	DefaultCommand  string `yaml:"default-command"`
+	DefaultHost     string  `yaml:"default-host"`
+	DefaultUser     string  `yaml:"default-user"`
+	DefaultPassword *string `yaml:"default-password"`
+	DefaultCommand  string  `yaml:"default-command"`
 	Subscribe       []struct {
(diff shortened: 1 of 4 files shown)


binwiederhier:
Love it!

PR #447, opened 2022-10-21
Allow reasding subscribe credentials from NTFY_USER env

SuperSandro2000:
This allows running ```ntfy sub test``` and having the credentials in an env. Tested locally with my ntfy server.


binwiederhier:
Thanks. Sorry it took so long.

PR #451, opened 2022-10-23
Remove buffering from nginx config, make config secure by default

SuperSandro2000:
Turning off proxy buffering is not recommend by upstream https://www.nginx.com/blog/avoiding-top-10-nginx-configuration-mistakes/#proxy_buffering-off by default. And making configuration more secure by removing TLSv1 TLSv1.1 and redirecting to https all the time to never leak credentials.

PS: https is not annoying and curl can follow redirects with -L.

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -448,31 +448,16 @@ or the root domain:
       server_name ntfy.sh;
 
       location / {
-        # Redirect HTTP to HTTPS, but only for GET topic addresses, since we want 
-        # it to work with curl without the annoying https:// prefix
-        set $redirect_https "";
-        if ($request_method = GET) {
-          set $redirect_https "yes";
-        }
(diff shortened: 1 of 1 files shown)


binwiederhier:
My apologies for not getting to this earlier, but I'm not sure I 100% agree with the PR. 

- I do agree with the TLS cipher suite changes, those should be in the example
- As I said before, I do want people to be able to use `curl` without typing the `https://` part, and I want the sample setup to reflect that. So maybe we can just have two examples like that? Let me try to see what I can whip up.

binwiederhier:
Added two nginx configs: One called "convienient" and one called "more secure" (yours). https://github.com/binwiederhier/ntfy/blob/main/docs/config.md?plain=1#L437

Also added some explanations. I don't have the time to try the output buffering stuff, but I swear I had issues with that.

PR #462, opened 2022-11-02
Add info for self-hosting on NixOS.

wamserma:


binwiederhier:
I don't know what I'm looking at, but it looks pretty harmless. Thanks for your contribution.

PR #469, opened 2022-11-07
Fix bug where GET or HEAD action requests could not be made from the web client

### LINKED ISSUE #468: A GET/HEAD action with no body does not send from the web client
While the backend properly validates that you cannot attach a body to a GET/HEAD request, the web client does not allow you to perform these actions, spitting this into the console

```
TypeError: Window.fetch: HEAD or GET Request cannot have a body.
```

MRE:
```
curl ntfy.sh \
  -d '{
    "topic": "5cba1ce1-6373-458a-83f5-105fe5d258d5",
    "message": "",
    "actions": [
      {
        "action": "http",
        "label": "Call it",
        "url": "http://127.0.0.1:8000",
        "method": "GET"
      }
    ]
  }'
```

ollien:
See #468 for details on the bug this fixes.

The cause of this comes down to a subtlety of the `fetch` API. If `body` is anything that is non-nullish, the browser will spit back that a body is not allowed for a GET/HEAD request. The fix is just to remove the null coalesce :)

Closes #468

### DIFF
--- a/web/src/components/Notifications.js
+++ b/web/src/components/Notifications.js
@@ -399,7 +399,9 @@ const performHttpAction = async (notification, action) => {
         const response = await fetch(action.url, {
             method: action.method ?? "POST",
             headers: action.headers ?? {},
-            body: action.body ?? ""
+            // This must not null-coalesce to a non nullish value. Otherwise, the fetch API
+            // will reject it for "having a body"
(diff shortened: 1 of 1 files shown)


binwiederhier:
Much appreciated. Nice catch!

PR #470, opened 2022-11-07
docs: fix addr-prefix type

snh:
Noticed this minor mistake in the documentation when setting up a self-hosted instance earlier.


binwiederhier:
So @snh, shameless question: I'm looking for a Principal/Staff Engineer position and GitHub sounds like a fun place to work. Are there any positions you could refer me to? :grimacing: (I'm sorry, I have no shame)

PR #499, opened 2022-11-14
docs: add UnifiedPush section

### LINKED ISSUE #497: UnifiedPush restrictions
A few things which are currently not covered by the documentation:

1. Does UP still work if i restrict the access by default? https://docs.ntfy.sh/config/#example-private-instance
2. What happens if i change the displayname of an active UP subscription in the app? e.g `https://ntfy.example.com/xxxxxxxc` to `tusky` ?

bt90:
Fixes #497

Added an ACL section which covers UnifiedPush.

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -309,6 +309,22 @@ with the given username/password. Be sure to use HTTPS to avoid eavesdropping an
     ]));
     ```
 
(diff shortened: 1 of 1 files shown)


binwiederhier:
Thanks!

PR #501, opened 2022-11-15
Fix the Kubernetes ConfigMap

QJoly:
Hi 

While migrating from docker (standalone) to Kubernetes, I noticed an indentation error on the config-map (causing a fatal error). 
This "huge" PR is fixing this problem :smile: 

thank you for your amazing project :)


binwiederhier:
Whether spaces or tabs, indentation matters. Thanks for the PR.

image

PR #502, opened 2022-11-15
Batch message INSERTs

binwiederhier:


binwiederhier:
> Is the new logging messages for every success/failure of adding + processing + pruning every single message in the queue going to bloat your logs?

Anything below INFO logs are not logged by default, only when log-level is DEBUG or TRACE. So I can turn them on and diagnose a little and then turn it off again. The log-level is the only option that can be hot-reloaded from the config without restarting the server. Makes debugging much easier: https://docs.ntfy.sh/config/#debuggingtracing

PR #533, opened 2022-12-08
Add "Generate topic name" button to "Subscribe to topic" dialog

### LINKED ISSUE #453: Add "Generate topic name" button
The web app and documentation should advocate for all topics that aren't "well known names" to have high entropy components. Common cryptographic wisdom is that a 2^-32 event is "not going to happen", so by the birthday bound the appropriate level of entropy is `(32 + log2(num_topics_created)) * 2` bits.

64 bits / 6 bits per base64 character = 10.6 characters of base64 (minimum recommendation: 11)
96 bits / 6 bits per base64 character = 16 characters of base64 - sufficient for servers with 65k random topics
128 bits / 6 bits per base64 character = 21.3 characters of base64 (reasonable maximum recommendation: 21)

 - The web app "Subscribe to topic" dialog should include a button to append a hyphen and 16 base64 characters to what's already typed in the dialog box.
 - https://ntfy.sh/docs/publish/ should say "Pick something with about 96 bits of entropy (16 base64 characters)" instead of "that's not easily guessable" and an interactive component to generate one in the browser.

yardenshoham:
Added a new button. When clicked it'll generate a random alphanumeric string and set it as the topic name.

* Closes #453

# Screenshot
image

# Video
https://user-images.githubusercontent.com/20454870/206441009-2e8098cc-c62d-4e2d-b5e5-95c4578988be.mp4

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -19,6 +19,7 @@
     "i18next-browser-languagedetector": "^6.1.4",
     "i18next-http-backend": "^1.4.0",
     "js-base64": "^3.7.2",
+    "nanoid": "^4.0.0",
     "react": "latest",
     "react-dom": "latest",
     "react-i18next": "^11.16.2",
--- a/web/package.json
+++ b/web/package.json
@@ -19,7 +19,6 @@
     "i18next-browser-languagedetector": "^6.1.4",
     "i18next-http-backend": "^1.4.0",
     "js-base64": "^3.7.2",
-    "nanoid": "^4.0.0",
     "react": "latest",
     "react-dom": "latest",
     "react-i18next": "^11.16.2",
--- a/web/src/components/SubscribeDialog.js
+++ b/web/src/components/SubscribeDialog.js
@@ -6,7 +6,7 @@ import Dialog from '@mui/material/Dialog';
 import DialogContent from '@mui/material/DialogContent';
 import DialogContentText from '@mui/material/DialogContentText';
 import DialogTitle from '@mui/material/DialogTitle';
-import {Autocomplete, Checkbox, FormControlLabel, useMediaQuery} from "@mui/material";
+import {Autocomplete, Checkbox, FormControlLabel, Grid, useMediaQuery} from "@mui/material";
 import theme from "./theme";
 import api from "../app/Api";
 import {topicUrl, validTopic, validUrl} from "../app/utils";
@@ -50,6 +50,14 @@ const SubscribeDialog = (props) => {
                 onSuccess={handleSuccess}
             />}
         </Dialog>
+    );
+};
+
+const Row = (props) => {
+    return (
+        <div style={{display: 'flex'}} role="row">
+            {props.children}
+        </div>
     );
 };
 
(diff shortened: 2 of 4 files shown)


web/package.json
```diff
@@ -19,6 +19,7 @@
     "i18next-browser-languagedetector": "^6.1.4",
     "i18next-http-backend": "^1.4.0",
     "js-base64": "^3.7.2",
+    "nanoid": "^4.0.0",
```
binwiederhier, line 22: Adding a new library for generating a random string is insane :-) That can easily be done in JS without a new lib.

web/src/components/SubscribeDialog.js
```diff
@@ -118,7 +131,8 @@ const SubscribePage = (props) => {
                         maxLength: 64,
                         "aria-label": t("subscribe_dialog_subscribe_topic_placeholder")
                     }}
-                />
+                    />
+                <Button onClick={generateTopicName} disabled={props.topic.includes("-")}>{t("subscribe_dialog_subscribe_button_generate_topic_name")}</Button><br />
```
binwiederhier, line 135: The button should in the same line as the text field, IMHO.

binwiederhier (changes requested):
Thank you for your contribution! Next time, please ask _before_ writing code if I'd accept it. I don't want people to write code and then have me reject it. This change is small enough so it's fine.

I left some comments. Most prominently, adding libraries is evil, and so for something as trivial as generating a random string we should just do in vanilla JS, or with the libs we already have.

web/src/components/SubscribeDialog.js
yardenshoham: Sure, done

web/package.json
yardenshoham: Dropped the dependency

web/src/components/SubscribeDialog.js
```diff
@@ -77,6 +89,17 @@ const SubscribePage = (props) => {
...
         props.onSuccess();
     };
+    const generateTopicName = () => {
+        const entropy = randomAlphanumericString();
+        let newTopic = props.topic;
+        if (newTopic) {
+            newTopic += "-" + entropy;
+        } else {
+            const sliceIndex = entropy.length / 2;
+            newTopic = entropy.slice(0, sliceIndex) + "-" + entropy.slice(sliceIndex);
```
binwiederhier, line 99: That seems quite random. Why the dash?

binwiederhier (approved):
Looks good

web/src/components/SubscribeDialog.js
binwiederhier: Okay I see it now. Didn't see the `contains(-)` before. 

That's ... unexpected and quite the weird UX. IMHO the "append with a dash" logic is also unexpected. A "Generate" button usually just overwrites the existing thing. And then you don't have to ever disable the button. Just make the button generate a random string. Easy as pie.

binwiederhier:
Awesome thanks

wunter8:
Sorry that in late to the party, but aren't topic names case-sensitive? So the alphabet could (should?) include a-z and A-Z? What about `_` and `-`? I know it doesn't have to be able to produce the full range of valid topics, but I figured I'd comment in case it wasn't considered.

Also, I think it might be good to include a sanity check that the generated topic doesn't start with `up`

yardenshoham:
> Sorry that in late to the party, but aren't topic names case-sensitive? So the alphabet could (should?) include a-z and A-Z? 

If they are case-sensitive, I guess we could add `A-Z` to the alphabet.

> What about `_` and `-`? I know it doesn't have to be able to produce the full range of valid topics, but I figured I'd comment in case it wasn't considered.

What are you suggesting here? Simply adding `_` to the alphabet?

> Also, I think it might be good to include a sanity check that the generated topic doesn't start with `up`

Why?

-----

Generally, I think going for the simplest thing is best

wunter8:
> What are you suggesting here? Simply adding `_` to the alphabet?

`_` and `-` are both valid topic characters, so they could be included in the alphabet as well, yes. 

> > Also, I think it might be good to include a sanity check that the generated topic doesn't start with `up`
> 
> Why?

All topics generated by/for UnifiedPush start with `up`. A user generated topic that starts with `up` acts like normal, but many users add anonymous-write permissions to `up*` topics. So if a randomly generated (and presumably secure) topic is generated that starts with `up`, it might be exposed to (malicious) anonymous writes. They would need to get hold of the randomly generated topic, so it's not likely anonymous writes to the topic would happen frequently, but it's a possibility.

This sanity check certainly isn't a necessity. It would just be out of an abundance of caution

PR #543, opened 2022-12-12
Accumulate incoming messages in a buffered channel

nicois:
Instead of using a deque, store incoming messages in a native buffered channel, if buffering is enabled.

In addition, modify the batching algorithm so the enforced delay between consecutive `addMessages` invocations is applied after all pending messages are processed. This acts as a "cooldown", rather than a "warmup". This avoids the need for more complex timing logic to dispatch batches, removes latency in adding messages when received infrequently, and natively blocking the goroutine until messages are received.

Because the message processing loop always performs a blocking read first, it is appropriate for low-throughput environments just as much as high-throughput ones.

The default value of batchSize has been changed to 10, with a zero cooldown. This means that when messages are arriving faster than they can be inserted into sqlite, they will automatically become batched in groups of up to 10. It also means requests will complete more quickly, slowing down only when the buffer fills - at which point they will block for a short period, similar to legacy behaviour.


nicois:
I haven't tried to update documentation or anything else, pending feedback. If reviews are positive I can tidy this up.

nicois:
Benefits of this change:
- requests are guaranteed to not block, as long as the buffered channel has capacity. (The Queue, which it replaces, would use an unbuffered channel to dequeue batches of messages, and each message would be executed by the sqlite client before the next one was processed)
-  in low-volume environments, when a message is received, it will be instantly flushed to the database; only when multiple messages arrive within the cooldown period will there be any delay. This also means that the default configuration values could be modified to always use a buffer, simplifying the code further.

Limitation of this change:
- If very large volumes of messages arrive during the cooldown period - sufficient to saturate the buffered channel - the request will block. The PR does not currently address this, but could be extended to do so if required/desired (or another follow-up PR could do so). However, this backpressure is aguably a good thing: if the system is receiving so many messages, it is probably preferable to act in this way. Some requests will pause for a short period before the channel has capacity (if the TCP server has accepted their connection), while if there are still more requests, they will be able to complete the TCP handshake, and the client can retry.

binwiederhier:
I haven't forgotten about this. My brain is fried. Sorry. Will review tomorrow I hope.

nicois:
I'm going to make a few more changes to this today, which IMO will simplify it further, so feel free to unfry your brain for another day or two.

nicois:
Added an extra test to exercise the queue buffering a little better, also improving coverage.

PR #549, opened 2022-12-17
Use pre-commit

nicois:
Add pre-commit hooks

    This will ensure that there is standard linting applied to files in
    the repository. It assists in avoiding conflicts due to different
    contributers' IDEs.

    It is recommended that contributers run the following:
    `pre-commit install -t pre-push -t pre-commit`

Without this, either I need to disable my IDE's auto-linting/fixing, or keep undoing the fixes it makes before pushing commits. Other contributers will be in a similar situation. Otherwise, their PRs will include syntax changes which distract from reviewing the real behavioural changes.

There are standard formatters for most file formats. Along with stripping trailing whitespace, this cleans up the files without modifying their behaviour in any way.


binwiederhier:
Soo I'm not a fan of this. Generally yes, but this is a little odd.

- I already have linting and gofmt in place for Go. 
- This formatter changes things that are just a little different, but not necessarily "correct". I believe it would even conflict with `gofmt`.
- The markdown cannot be formatted because it's not standard markdown (which doesn't exist anyway). It's an mkdocs dialect, so if you look closely, you destroyed some of the formatting with this.

nicois:
> Soo I'm not a fan of this. Generally yes, but this is a little odd.
> 
> * I already have linting and gofmt in place for Go.

pre-commit is the standard way of making sure these things are applied. Being git hooks, they are automatically run each time, helping to ensure what is pushed is good code.

> * This formatter changes things that are just a little different, but not necessarily "correct". I believe it would even conflict with `gofmt`.

`https://github.com/mvdan/gofumpt` is a stricter validator that `gofmt`, while being backwards-compatible. It auto-fixes any problems, but since you're not a fan, I've removed it.

> * The markdown cannot be formatted because it's not standard markdown (which doesn't exist anyway). It's an mkdocs dialect, so if you look closely, you destroyed some of the formatting with this.
I've instructed `prettier` to not modify markdown files.

The other hooks are also valuable. There are many files with trailing whitespace, causing spurious line changes when someone touches a file, for example.

nicois:
I do recommend reviewing the additional autofixes includes in `gofumpt`. IMO they improve code readability:

https://github.com/mvdan/gofumpt#added-rules

binwiederhier:
Thank you for your work. However, I don't think I care much for this, and I'd rather focus my energy elsewhere. I may revisit this later, though.

Issue #552, opened 2022-12-21
Enable HTTP/2 for ntfy.sh


otbutz:
Any reason that https://ntfy.sh is limited to HTTP/1.1 ?

binwiederhier:
Good question. No particular reason, other than I have never set that up for nginx. I could probably do it. Do you happen to have experience with that @otbutz?

Here's how it's set up: https://github.com/binwiederhier/ntfy-ansible/blob/main/roles/nginx/templates/sites-available/ntfy.j2

binwiederhier:
Looks like it's just adding the `http2` stanza: https://ubiq.co/tech-blog/how-to-enable-http2-in-nginx/

Maybe I'll yolo-deploy that tomorrow and see what happens.

otbutz:
I would follow Mozillas best practices and also enable TLS1.3 while you're at it:

https://ssl-config.mozilla.org/#server=nginx&version=1.17.7&config=intermediate&openssl=1.1.1k&hsts=false&ocsp=false&guideline=5.6 (You should really consider to ditch HTTP and enforce HTTPS with HSTS.)

You could also switch to caddy if you're feeling adventurous and get HTTP/3 support :wink:

binwiederhier:
> I would follow Mozillas best practices and also enable TLS1.3 while you're at it

Sure why not.

> You should really consider to ditch HTTP and enforce HTTPS with HSTS

Nope. That prevents the awesome `curl -d hi ntfy.sh/yo` usage. Don't wanna type `https://` all the time, and I am usually not fearing MITM or sharing state secrets. If people are, they can set up their own server.

> You could also switch to caddy if you're feeling adventurous and get HTTP/3 support wink

If it ain't broke, don't fix it. :-D

otbutz:
That should still work without https:// if you allow curl to follow redirects:

```sh
curl -L -d hi ntfy.sh/yo
```

binwiederhier:
See https://github.com/binwiederhier/ntfy-ansible/pull/1

binwiederhier:
Sadly the Unix socket change causes a (benign) warning in the ntfy code for every single request (`r.RemoteAddr` now contains `@`, instead of an IP address), which means I need a small code change to handle that case. Otherwise I'd get flooded in warnings.

otbutz:
Shouldn't the remote address not be populated from `X-Forwarded-For` header in this case?

Issue #519, opened 2022-11-26
IPv6 on the official ntfy.sh server


bernhardschmidt:
I would highly appreciate it if you would enable IPv6 connectivity on the official ntfy.sh service. Right now it seems to be hosted on DigitalOcean. Enabling IPv6 there should be pretty straight forward.

https://docs.digitalocean.com/products/networking/ipv6/

binwiederhier:
There are multiple reasons I haven't done that yet:

1. I know nothing about IPv6
2. There are assumptions in the server about IPv4 addresses (for rate limiting mostly)
3. Rate limiting is 100% based around IPv4 addresses right now. Given that IPv6 addresses are cheap I think I'd be opening myself up to some bad times.

So while intriguing, I would not hold my breadth right now. I'm always happy to accept code contributions or discuss it further though.

gsauthof:
When enabling IPv6, you could rate limit IPv6 addresses using only their `/48`  prefix, as this the smallest prefix usually assigned to an entity. So this should be pretty conservative approach, for  a start.

However, most entities surely just get `/56` or even `/64` prefixes, so a less overly cautious approach would be to rate-limit those.

Or use a tiered approach were you have an IPv4-like rate limit for `/64` prefixes and higher limits for `/56` and then an even higher one for `/48`.

Background: With IPv6, the address is 128 bit wide, and the idea is that a network gets at least a `/64` prefix delegated. If a provider hands out a `/48` prefix that means that the customer may create up to `2**16` networks under that prefix, i.e. he/she is able to sub-delegate longer prefixes in his/her own infrastructure quite flexible.

binwiederhier:
Using a large-enough prefix sounds like a reasonable compromise. It's still not high on my list, though; so much other stuff to do :-D

arjan-s:
I think there may be more to do than just enabling IPv6 on the main server. My personal server has both IPv4 and IPv6, my home has both too, but my mobile provider only has IPv4. I've noticed that the ntfy Android client regularly loses connection to my dual-stack ntfy instance and takes a very long time to reconnect, which causes my notifications to arrive very late and in batches. My second ntfy instance only has an IPv4 DNS record and I have none of the problems there.

binwiederhier:
This is a great article that describes the reasons for why ntfy.sh does not support IPv6 yet: https://adam-p.ca/blog/2022/02/ipv6-rate-limiting/

beedell-roke:
[...]

PR #553, opened 2022-12-22
docs: improve nginx config

bt90:
Incorporates the changes mentioned in https://github.com/binwiederhier/ntfy/issues/552:

* enable IPv6
* enable HTTP2
* ssl config according to https://ssl-config.mozilla.org/#server=nginx&version=1.17.7&config=intermediate&openssl=1.1.1k&hsts=false&ocsp=false&guideline=5.6

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -471,6 +471,7 @@ or the root domain:
 
     server {
       listen 80;
+      listen [::]:80;
       server_name ntfy.sh;
 
       location / {
@@ -554,6 +555,7 @@ or the root domain:
     
     server {
       listen 80;
+      listen [::]:80;
       server_name ntfy.sh;
 
       location / {
--- a/docs/config.md
+++ b/docs/config.md
@@ -508,13 +508,17 @@ or the root domain:
     }
     
     server {
-      listen 443 ssl;
+      listen 443 ssl http2;
+      listen [::]:443 ssl http2;
       server_name ntfy.sh;
     
       ssl_session_cache builtin:1000 shared:SSL:10m;
+      ssl_session_timeout 1d;
+      ssl_session_tickets off;
+      
       ssl_protocols TLSv1.2 TLSv1.3;
-      ssl_ciphers HIGH:!aNULL:!eNULL:!EXPORT:!CAMELLIA:!DES:!MD5:!PSK:!RC4;
-      ssl_prefer_server_ciphers on;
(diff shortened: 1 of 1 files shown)


binwiederhier:
Thanks for the PR. I actually tried the http2 config in a VM and it seems fine. Though the proxy_http_version is still set to 1.1. Gotta try that too. The real production config is here btw: https://github.com/binwiederhier/ntfy-ansible/ -- What you changed is just the docs (which is also important!).

As for IPv6, I don't think I can enable that without changes to the application. See #519 for details. So maybe lets split those changes into two separate things.

binwiederhier:
That mozilla SSL config tool is amazing!

bt90:
So far no problems with IPv6, although I use caddy as a reverse proxy. At least for the self-hosted usecase, rate limiting should not be a huge problem. But I agree that it's tricky: https://adam-p.ca/blog/2022/02/ipv6-rate-limiting/

Applying rate limits to, say, /64 subnets seems like a solid approach.

btw sorry for the confusion. @otbutz is my work account ;)

bt90:
> Though the proxy_http_version is still set to 1.1

Without TLS in the backend that's most you can get if you don't support h2c (HTTP2 cleartext):

https://pkg.go.dev/golang.org/x/net/http2/h2c

The benefit is most likely not noticeable. HTTP/2 and HTTP/3 only offer real improvements over a WAN link.

bt90:
A Unix domain socket on the other hand might be a real improvement 😉

binwiederhier:
> A Unix domain socket on the other hand might be a real improvement wink

ntfy does support unix sockets. Actually that may be pretty awesome, because that would cut the number of open files in half, right?

binwiederhier:
See https://github.com/binwiederhier/ntfy-ansible/pull/1

francislavoie:
FYI, the Mozilla SSL config site is no longer maintained, as far as I know. https://twitter.com/CubicleApril/status/1438940565916225545

I would suggest using Caddy instead to simplify the setup here. Disclaimer: I help maintain Caddy. A Caddyfile config would be as simple as this, I believe:

```
ntfy.my.domain {
    reverse_proxy ntfy:80
}
```

Automates HTTPS, HTTP->HTTPS redirects, HTTP/2 and HTTP/3, modern security by default, good proxy defaults. Does everything that Nginx config does, in only 3 lines of config, and removes the need for certbot.

bt90:
@binwiederhier should I remove the IPv6 related changes for now?

binwiederhier:
Hi @francislavoie, I have heard only good things about Caddy. I look forward to the day that I get to use it. I have instructions for Caddy in the docs already, but I have not (yet) switched on ntfy.sh. 

Right now I have no reason to switch, because everything works beautifully and I have 8,000,000 other things to work on :-) While I'm sure that Caddy is easy, it's another thing to learn and I'd spend a few days changing and testing the ansible playbook and such. _If it ain't broke, don't fix it..._ There will come a day when it'll be Caddy's turn!

@bt90 Let me take the rest of the PR and change it according to what I've done in the ansible. You've done enough. Thank you for kicking this off (despite the fact that it caused a bunch of work that I didn't want to do, hehe). The nginx config was quite outdated.

PR #555, opened 2022-12-23
docker: add basic healthcheck

bt90:
@binwiederhier  couldn't spot any kind of version endpoint, or I would have preferred that.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,5 +3,7 @@ MAINTAINER Philipp C. Heckel <philipp.heckel@gmail.com>
 
 COPY ntfy /usr/bin
 
+HEALTHCHECK --interval=60s --timeout=10s CMD wget --no-verbose --tries=1 --spider http://localhost/config.js || exit 1
+
 EXPOSE 80/tcp
 ENTRYPOINT ["ntfy"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@ MAINTAINER Philipp C. Heckel <philipp.heckel@gmail.com>
 
 COPY ntfy /usr/bin
 
-HEALTHCHECK --interval=60s --timeout=10s CMD wget --no-verbose --tries=1 --spider http://localhost/config.js || exit 1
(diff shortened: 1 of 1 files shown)


binwiederhier:
I added a preliminary health API endpoint for this. Seems like a good thing to have. For now it's just a dummy endpoint that returns 200, but we could make that more functional later: https://github.com/binwiederhier/ntfy/commit/dd282963c34dff17d37c2464fdd95318327f74c0

```
curl -v localhost:2586/v1/health
*   Trying 127.0.0.1:2586...
* Connected to localhost (127.0.0.1) port 2586 (#0)
> GET /v1/health HTTP/1.1
> Host: localhost:2586
> User-Agent: curl/7.81.0
> Accept: */*
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< Access-Control-Allow-Origin: *
< Content-Type: text/json
< Date: Sat, 24 Dec 2022 17:25:05 GMT
< Content-Length: 17
< 
{"healthy":true}
* Connection #0 to host localhost left intact
```

bt90:
Should the `/v1` be blacklisted in `config.js` ?

binwiederhier:
The disallowed topics are just for paths that do not have a `/`, so `v1/health` is fine :-)

bt90:
@binwiederhier adapted the check for the `v1/health` endpoint.

binwiederhier:
Do you think the grep is even necessary right now? I mean if it returns a non-200 code, wget will fail and retry anyway, right?

bt90:
Depends on the API. If a `"health":false` is guaranteed to be sent with a non-200 code, we can safely drop the grep.

It's just a bit odd as you're querying the health of the server and the status code is for the query. If the API would return `"health":false` because your server is burning, it would be a valid response for the query and i would expect a HTTP 200.

binwiederhier:
> and i would expect a HTTP 200

I honestly don't know what the best practice is here. I tried to Google but without success. Let's just leave it like this for now.

PR #562, opened 2022-12-30
Added clarification on client.yml configuration

fleopaulD:
I didn't understand why the `ntfy publish --debug topic message` command don't choose the default-host I entered in `/etc/ntfy/client.yml`. 
If command is run as root -> config file = `/etc/ntfy/client.yml` 
If command is run as non-root -> config file = `~/.config/ntfy/client.yml` 
I think this is an important precision for users.


binwiederhier:
This seems like a no-brainer change. Thank you. It has sadly caused quite some confusion with users, even though it is standard practice.

PR #582, opened 2023-01-13
Fix small issues in the K8s sts documentation

Remedan:
When I was deploying ntfy to my Kubernetes cluster, I noticed two small mistakes in the documentation.

The flag `--cache-file` and its argument need to be passed as two separate arguments; otherwise it gets parsed as a single long flag and results in an "incorrect usage" error.

The pvc needs to be mounted to actually get used.


binwiederhier:
Awesome, thank you.

PR #584, opened 2023-01-17
Rough draft of reciever rate limiting for discussion

karmanyaahm:
This is the original proposal from the Matrix room
me:
> My main idea, the ideal solution (as PeterCxy proposed) is to move the cost, for at least UP rate limits, to the receiver instead of the sender.
> 
> There are many ways to do it, but one potential solution (a simple but clean-ish one I could think of):
> 
>  -   Like id: and user:, make up* topics a up: visitor. [1]
>  -  This up: visitor (the UP topic) has large-ish rate limits (to prevent obvious spam/misconfiguration), but doesn't care too much.
>  -  Topic.Subscribe and Topic.Unsubscribe keep track of the most recent visitor to subscribe to a topic, and the time (say prev. 12hrs is valid). [2]
>  -   On Topic.Publish, the message is 'billed' to the most-recent-visitor's rate limit (either ip: or user:), rejected if there is no recent visitor.
>  -   Delete the message if successfully delivered to at least one client. UP topics are unique to each client anyways. [3]
> 
> [1] In which case up* topics have to be reserved explicitly for receiver-based rate limiting. We cannot rely on up=1 because the client doesn't control that.
> [2] If all of this is in-memory: say someone loses their connection, wait 1 hour, ntfy restarts, wait 1 hour, they resubscribe. If there was a notification at T+90 minutes, that notification would be lost, since the topic would show up as 'fresh', and the notification rejected. However, this seems like a fair compromise for the simplicity of avoiding a database.
> [3] Attempt delivering to multiple clients if they are online and subscribed though, for debugging purposes. This is just an extra resource-saving measure and not part of the core plan.
> 
> I can actually help you make it now, though, once we agree on a plan, since this is needed by a clear deadline.

binwiederhier:
> I think I understand this plan. In my own words:
> 
> For topics starting with up*, we keep track of the visitors that are subscribed to a topic in the topic struct.
> 
> When a message is published, we determine the visitor based on the topic name instead of the IP (v := s.visitorByTopic("upABCDEF..")). If there are multiple visitors, we pick the most recent one.
> 
> The rest of the logic remains the same.
> 
> Correct?

me:
> Basically, yes. The goal is to 'bill' that messsage to the most recently subscribed visitor instead of to the sender.
> 
> Additionally, this would require formally defining 'up*' as a special namespace, since messages are "billed" differently.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -137,12 +137,14 @@ var (
 )
 
 const (
-	firebaseControlTopic     = "~control"                // See Android if changed
-	firebasePollTopic        = "~poll"                   // See iOS if changed
-	emptyMessageBody         = "triggered"               // Used if message body is empty
-	newMessageBody           = "New message"             // Used in poll requests as generic message
-	defaultAttachmentMessage = "You received a file: %s" // Used if message body is empty, and there is an attachment
-	encodingBase64           = "base64"
+	firebaseControlTopic            = "~control"                // See Android if changed
+	firebasePollTopic               = "~poll"                   // See iOS if changed
+	emptyMessageBody                = "triggered"               // Used if message body is empty
+	newMessageBody                  = "New message"             // Used in poll requests as generic message
+	defaultAttachmentMessage        = "You received a file: %s" // Used if message body is empty, and there is an attachment
+	encodingBase64                  = "base64"
+	unifiedpushTopicPrefix          = "up"
+	unifiedPushSubscriptionDuration = 12 * time.Hour
 )
 
 // WebSocket constants
@@ -544,6 +546,17 @@ func (s *Server) handlePublishWithoutResponse(r *http.Request, v *visitor) (*mes
 	if err != nil {
 		return nil, err
 	}
+
+	var v_billing *visitor
+	if strings.HasPrefix(t.ID, unifiedpushTopicPrefix) {
+		v_billing := t.getBillee()
+		if v_billing != nil {
+			// instant reject and won't even store it if there's no one registered for a UP topic in the past some time
+	// need to find error code for device not available try again later
+			return nil, errHTTPInternalError
+		}
+	}
+
 	if err := v.MessageAllowed(); err != nil {
 		return nil, errHTTPTooManyRequestsLimitMessages
 	}
@@ -569,6 +582,7 @@ func (s *Server) handlePublishWithoutResponse(r *http.Request, v *visitor) (*mes
 	if m.Message == "" {
 		m.Message = emptyMessageBody
 	}
+	// we do not need to handle delays, because 1. for UP delays are not needed 2. if the up app server is adding a delay it is shooting itself in the foot
 	delayed := m.Time > time.Now().Unix()
 	log.Debug("%s Received message: event=%s, user=%s, body=%d byte(s), delayed=%t, firebase=%t, cache=%t, up=%t, email=%s",
 		logMessagePrefix(v, m), m.Event, m.User, len(m.Message), delayed, firebase, cache, unifiedpush, email)
@@ -582,6 +596,7 @@ func (s *Server) handlePublishWithoutResponse(r *http.Request, v *visitor) (*mes
 		if s.firebaseClient != nil && firebase {
 			go s.sendToFirebase(v, m)
 		}
+		// same as delays, it should count against app servers
 		if s.smtpSender != nil && email != "" {
 			v.IncrementEmails()
 			go s.sendEmail(v, m, email)
@@ -598,7 +613,11 @@ func (s *Server) handlePublishWithoutResponse(r *http.Request, v *visitor) (*mes
 			return nil, err
 		}
 	}
-	v.IncrementMessages()
+	if v_billing != nil {
+		v_billing.IncrementMessages()
+	} else {
+		v.IncrementMessages()
+	}
 	if s.userManager != nil && v.user != nil {
 		s.userManager.EnqueueStats(v.user)
 	}
@@ -961,7 +980,7 @@ func (s *Server) handleSubscribeHTTP(w http.ResponseWriter, r *http.Request, v *
 	}
 	subscriberIDs := make([]int, 0)
 	for _, t := range topics {
-		subscriberIDs = append(subscriberIDs, t.Subscribe(sub))
+		subscriberIDs = append(subscriberIDs, t.Subscribe(sub, v))
 	}
 	defer func() {
 		for i, subscriberID := range subscriberIDs {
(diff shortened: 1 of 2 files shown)


server/topic.go
```diff
@@ -1,43 +1,75 @@
...
+	for _, this_subscriber := range t.subscribers {
+		return this_subscriber.v
+	}
+	// what if someone unsubscribed and DOESNT want their sub to count against them anymore, maybe the app server lost sync and will keep on sending stuff
+	// I guess they suffer for unifiedPushSubscriptionDuration?
+
+	// if lastunsub v exists, and the time since it was unsubbed is longer than our limit, it should not exist
+	if t.lastUnsub.v != nil && time.Since(t.lastUnsub.unsubTime) > unifiedPushSubscriptionDuration {
+		t.lastUnsub.v = nil
+	}
```
karmanyaahm, line 51: That. How do we know whether the last subscriber wants to cache messages (gone temporarily) or not (unsubscribed permanently)?

server/server.go
```diff
@@ -544,6 +546,17 @@ func (s *Server) handlePublishWithoutResponse(r *http.Request, v *visitor) (*mes
...
+
+	var v_billing *visitor
+	if strings.HasPrefix(t.ID, unifiedpushTopicPrefix) {
+		v_billing := t.getBillee()
+		if v_billing != nil {
+			// instant reject and won't even store it if there's no one registered for a UP topic in the past some time
+	// need to find error code for device not available try again later
+			return nil, errHTTPInternalError
+		}
+	}
```
karmanyaahm, line 558: If there is no last subscriber, or there was a subscriber very long ago, what should ntfy do? 
1. Cache it anyway but bill it to the app server - inconsistent
2. Reject it. Which status code? 500?

server/topic.go
```diff
@@ -1,43 +1,75 @@
...
 // newTopic creates a new topic
 func newTopic(id string) *topic {
 	return &topic{
 		ID:          id,
-		subscribers: make(map[int]subscriber),
+		subscribers: make(map[int]topicSubscription),
 	}
 }
 
+// need a better name for bill?
```
karmanyaahm, line 36: Is the term "Bill" fine?

karmanyaahm: (apparently Billee is not a real word)

server/server.go
karmanyaahm: *should be == not !=

server/topic.go
MayeulC: Invoiced? Billed visitor/account?  Charged? Accounted?

Other "Bill" synonyms: cost, expense, tally

server/server.go
binwiederhier: I think this is kind of up to the UP spec, no? "What should the push server do if there is no one listening" seems like a spec question.

Obviously the easiest thing would be to reject it. Everything else would cause lots of headscratching.

The response code may be https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/412?

server/server.go
```diff
@@ -598,7 +613,11 @@ func (s *Server) handlePublishWithoutResponse(r *http.Request, v *visitor) (*mes
 			return nil, err
 		}
 	}
-	v.IncrementMessages()
+	if v_billing != nil {
+		v_billing.IncrementMessages()
+	} else {
+		v.IncrementMessages()
+	}
```
binwiederhier, line 620: I would generally opt for replacing `v` early in this function, instead of adding if-then-elses everywhere. 

Case and point: There's a `v.MessageAllowed()` at the top which would be run against the IP-visitor and not the subscriber visitor.

server/topic.go
binwiederhier: `t.SubscriberVisitor()`?

server/topic.go
binwiederhier: I don't understand what's happening here. Sorry.

server/server.go
```diff
@@ -544,6 +546,17 @@ func (s *Server) handlePublishWithoutResponse(r *http.Request, v *visitor) (*mes
 	if err != nil {
 		return nil, err
 	}
+
+	var v_billing *visitor
+	if strings.HasPrefix(t.ID, unifiedpushTopicPrefix) {
```
binwiederhier, line 551: Here's an idea: instead of using the prefix, could we use the presence of the `?up=1` query param? If the qurey param is present, bill differently? I vaguely remember this being discussed before, but I do not recall what the outcome was

karmanyaahm: From the UP dev chat

> up=1 is sent by the server, whereas these messages are being billed to the subscriber. I thought about that too, but then came up with this:
> 
>   -  I am subscribed to [ntfy.sh/videogameupdates](http://ntfy.sh/videogameupdates), an open channel for anyone to post
>   -  A malicious actor comes along and sends 10000 messages to [ntfy.sh/videogameupdates?up=1](http://ntfy.sh/videogameupdates?up=1)
>   - Me, poor subscriber, is now rate-limited 
> 
> So, we need a way in which the _subscriber knows that they are subscribing to a subscriber-billed topic_ instead of trusting the sender to inform us. The easiest way is the topic.

server/server.go
karmanyaahm: Good point

server/topic.go
karmanyaahm: Messages are being billed to the last visitor (either someone currently subscribed or the last person to unsubscribe). This bill-the-last-unsusbscriber behavior remains for `unifiedPushSubscriptionDuration`.

Let's say `user:KM` is subscribed to ntfy.sh/upABCD, and `unifiedPushSubscriptionDuration := 10 * time.Hour`

**Scenario 1:**
06:00 `user:KM`'s internet went out and they unsubscribed from ntfy.sh/upABCD
06:15: ntfy.sh receives a message at upABCD. It is stored and billed to `KM`
06:30 `user:KM` is online again. The missed message is delivered and everyone's happy.

**Scenario 2:**
06:00 `user:KM`'s internet went out and they unsubscribed from ntfy.sh/upABCD
06:15: ntfy.sh receives a message at upABCD. It is stored and billed to `KM`
15:59: ntfy.sh receives a message at upABCD. It is stored and billed to `KM`
16:01: ntfy.sh receives a message at upABCD. `t.lastUnsub.v` is set to `nil`. The message is rejected (412 or whatever else).
23:59 `user:KM` is online again. They missed some messages, but it's not *that* big of a deal.

**Scenario 3:**
06:00 `user:KM`'s intentionally unsubscribed from ntfy.sh/upABCD. They don't want upABCD messages anymore, but the app server keeps sending them.
06:15: ntfy.sh receives a message at upABCD. It is stored and billed to `KM`
15:59: ntfy.sh receives a message at upABCD. It is stored and billed to `KM`
16:01: ntfy.sh receives a message at upABCD. `t.lastUnsub.v` is set to `nil`. The message is rejected (412 or whatever else).
`user:KM` is angry that 2 (or 200 or 2000) messages from their rate limit were eaten up by this random app server.

server/server.go
karmanyaahm: Yeah, but in that case, logs and email-allowed need to be if-then-elsed (which makes more sense I guess, since there's fewer of those things). I'll figure out a way to minimize complications.

karmanyaahm:
Closing in favor of a new PR (coming soon)

PR #594, opened 2023-01-22
Elements requiring chown to run non-root Docker

jpbaril:
We also have to chown the attachments directory otherwise the docker container does not start and crashes. BTW, all that should be automated at the container creation. Because it took me at least an hour to understand that the only way to accomplish that chown command was to first launch the container as root, run the commands, and only then edit docker-compose.yml to add uid/gid. After that I could restart the container and it would now not crash.

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -287,7 +287,7 @@ services:
     restart: unless-stopped
 ```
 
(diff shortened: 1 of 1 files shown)


binwiederhier:
Thanks!

Issue #601, opened 2023-02-04
Authentication: match username from configurable header


ngerstle:
I'd like to elevate https://github.com/binwiederhier/ntfy/issues/19#issuecomment-1023118885  as an explicit feature request, that I believe enables administrators/users that self-host to handle both https://github.com/binwiederhier/ntfy/issues/530 and https://github.com/binwiederhier/ntfy/issues/296 (and may make a custom solution for https://github.com/binwiederhier/ntfy/issues/297 easier for users to implement).

I'd propose a configuration option (`user-header`) specifying a trusted header that a proxy in front of ntfy could set, and whose value is matched to a username. If the value matches an existing username, then ntfy assumes the request/connection to originate from that user, delegating the responsibility for credential validation to the proxy.

Example:
If the configuration for the server (`server.yml`) has the following: 
```
user-header: x-forwarded-user
```
Then a request reaching ntfy with
```
POST /mytopic HTTP/1.1
Host: ntfy.sh
x-forwarded-user: ngerstle

Backup successful
```
would be trusted to be made by the user `ngerstle`.

Authorization, and user management (excepting possible changes around passwords, if desired), would not change.

This would allow any number of different authentication options for access (depending on client support), ranging from the presently implemented basic, to LDAP, OIDC, OAuth2, JWT, mTLS, etc.
Options for proxies are numerous, and include (among others):
- [traefik](https://doc.traefik.io/traefik/middlewares/http/forwardauth/)
- [caddy](https://caddyserver.com/docs/caddyfile/directives/forward_auth)
- [oauth2-proxy](https://github.com/oauth2-proxy/oauth2-proxy)
- [pomerium](https://www.pomerium.com/)
- [cloudflare](https://developers.cloudflare.com/cloudflare-one/identity/)
- [tailscale](https://github.com/tailscale/tailscale/tree/main/cmd/nginx-auth)

If a dependency on a proxy is an acceptable requirement for authentication, then existing password storage and validation code could be removed entirely.

Client support for different authentication forms would vary, but differences already exist between client support for different features (streams, for example). The selection of authentication mechanism(s) can be left to administrators, and they would need to evaluate which clients support what forms of authentication. 
[...]

PR #602, opened 2023-02-04
Draft: user-header authentication

ngerstle:
This begins to sketch out a hacky way to implement https://github.com/binwiederhier/ntfy/issues/601.
Not being particularly familiar with go, or the ntfy code base, I haven't tried to compile or run this code.

I'm not particulary comfortable with the interface between `auth` and `server`- unless basic auth support is removed from `auth`, I'd suggest moving code for credential extraction from the request to the `auther`, whether that's using basic auth or a configured header, and select which auther to use during server setup.

### DIFF
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -8,6 +8,9 @@ import (
 
 // Auther is a generic interface to implement password-based authentication and authorization
 type Auther interface {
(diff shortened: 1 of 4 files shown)


binwiederhier:
I'm closing this, since it's very much out of date. I'll leave the ticket open since it's a good idea IMHO.

PR #609, opened 2023-02-14
Subscriber Billed Topics

karmanyaahm:
I am going with bill-to-last-visitor-for-12-hrs even though it doesn't [allow users to unsubscribe](https://github.com/binwiederhier/ntfy/pull/584#discussion_r1083381767) because it is the simplest to explain. It's much easier to document and explain _if you subscribe to a `up_` topic, you are responsible for whatever is sent there for the next 12 hrs, so keep it secret_, than anything else I could come up with.

The only other simple-for-the-user options would be: 
1. _You can only subscribe to 30 UP topics in a day. Each UP topic can receive 100 messages / day._ However, that requires keeping track of all those states and adds significant complexity to the code.
2. Bill in `sub := func(v *visitor, msg *message) error {`, so only messages you actually receive count against you. However, that allows spamming the cache with messages that no one is billed for, on the sending side.

I'll add tests and rebase it and stuff if you're good with this general implementation.

### DIFF
--- a/server/errors.go
+++ b/server/errors.go
@@ -3,8 +3,9 @@ package server
 import (
 	"encoding/json"
 	"fmt"
-	"heckel.io/ntfy/log"
 	"net/http"
+
+	"heckel.io/ntfy/log"
 )
 
 // errHTTP is a generic HTTP error for any non-200 HTTP error
@@ -92,5 +93,4 @@ var (
 	errHTTPInternalError                             = &errHTTP{50001, http.StatusInternalServerError, "internal server error", ""}
 	errHTTPInternalErrorInvalidPath                  = &errHTTP{50002, http.StatusInternalServerError, "internal server error: invalid path", ""}
 	errHTTPInternalErrorMissingBaseURL               = &errHTTP{50003, http.StatusInternalServerError, "internal server error: base-url must be be configured for this feature", "https://ntfy.sh/docs/config/"}
-	errHTTPWontStoreMessage                          = &errHTTP{50701, http.StatusInsufficientStorage, "topic is inactive; no device available to recieve message", ""}
 )
--- a/server/errors.go
+++ b/server/errors.go
@@ -92,4 +92,5 @@ var (
 	errHTTPInternalError                             = &errHTTP{50001, http.StatusInternalServerError, "internal server error", ""}
 	errHTTPInternalErrorInvalidPath                  = &errHTTP{50002, http.StatusInternalServerError, "internal server error: invalid path", ""}
 	errHTTPInternalErrorMissingBaseURL               = &errHTTP{50003, http.StatusInternalServerError, "internal server error: base-url must be be configured for this feature", "https://ntfy.sh/docs/config/"}
+	errHTTPWontStoreMessage                          = &errHTTP{50701, http.StatusInsufficientStorage, "topic is inactive; no device available to recieve message", ""}
 )
--- a/server/server.go
+++ b/server/server.go
@@ -1023,7 +1023,7 @@ func (s *Server) handleSubscribeHTTP(w http.ResponseWriter, r *http.Request, v *
 	defer cancel()
 	subscriberIDs := make([]int, 0)
 	for _, t := range topics {
-		subscriberIDs = append(subscriberIDs, t.Subscribe(sub, v.MaybeUserID(), cancel))
+		subscriberIDs = append(subscriberIDs, t.Subscribe(sub, v, cancel))
 	}
 	defer func() {
 		for i, subscriberID := range subscriberIDs {
@@ -1155,7 +1155,7 @@ func (s *Server) handleSubscribeWS(w http.ResponseWriter, r *http.Request, v *vi
 	}
 	subscriberIDs := make([]int, 0)
 	for _, t := range topics {
-		subscriberIDs = append(subscriberIDs, t.Subscribe(sub, v.MaybeUserID(), cancel))
+		subscriberIDs = append(subscriberIDs, t.Subscribe(sub, v, cancel))
 	}
 	defer func() {
 		for i, subscriberID := range subscriberIDs {
--- a/server/server.go
+++ b/server/server.go
@@ -372,6 +372,7 @@ func (s *Server) handleError(w http.ResponseWriter, r *http.Request, v *visitor,
 	}
 	w.Header().Set("Content-Type", "application/json")
 	w.Header().Set("Access-Control-Allow-Origin", s.config.AccessControlAllowOrigin) // CORS, allow cross-origin requests
+	w.Header().Set("TTL", "0")                                                       // if message is not being stored because of an error, tell them
 	w.WriteHeader(httpErr.HTTPCode)
 	io.WriteString(w, httpErr.JSON()+"\n")
 }
@@ -605,6 +606,14 @@ func (s *Server) handlePublishWithoutResponse(r *http.Request, v *visitor) (*mes
 	if err != nil {
 		return nil, err
 	}
+	v_old := v
+	if strings.HasPrefix(t.ID, subscriberBilledTopicPrefix) {
+		v = t.getBillee()
+		if v == nil {
+			return nil, errHTTPWontStoreMessage
+		}
+	}
+
 	if !v.MessageAllowed() {
 		return nil, errHTTPTooManyRequestsLimitMessages
 	}
@@ -639,8 +648,9 @@ func (s *Server) handlePublishWithoutResponse(r *http.Request, v *visitor) (*mes
 			"message_email":       email,
 		}).
 		Debug("Received message")
+		//Where should I log the original visitor vs the billing visitor
 	if log.IsTrace() {
-		logvrm(v, r, m).
+		logvrm(v_old, r, m).
 			Tag(tagPublish).
 			Field("message_body", util.MaybeMarshalJSON(m)).
 			Trace("Message body")
@@ -684,6 +694,10 @@ func (s *Server) handlePublish(w http.ResponseWriter, r *http.Request, v *visito
 	if err != nil {
 		return err
 	}
+
+	w.Header().Set("TTL", strconv.FormatInt(m.Expires-m.Time, 10)) // return how long a message will be stored for
+
+	// using m.Time, not time.Now() so the value isn't negative if the request is processed at a second boundary
 	return s.writeJSON(w, m)
 }
 
--- a/server/server.go
+++ b/server/server.go
@@ -437,13 +437,14 @@ func (s *Server) handleInternal(w http.ResponseWriter, r *http.Request, v *visit
 	} else if r.Method == http.MethodOptions {
 		return s.limitRequests(s.handleOptions)(w, r, v) // Should work even if the web app is not enabled, see #598
 	} else if (r.Method == http.MethodPut || r.Method == http.MethodPost) && r.URL.Path == "/" {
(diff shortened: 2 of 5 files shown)


server/errors.go
```diff
@@ -92,4 +92,5 @@ var (
 	errHTTPInternalError                             = &errHTTP{50001, http.StatusInternalServerError, "internal server error", ""}
 	errHTTPInternalErrorInvalidPath                  = &errHTTP{50002, http.StatusInternalServerError, "internal server error: invalid path", ""}
 	errHTTPInternalErrorMissingBaseURL               = &errHTTP{50003, http.StatusInternalServerError, "internal server error: base-url must be be configured for this feature", "https://ntfy.sh/docs/config/"}
+	errHTTPWontStoreMessage                          = &errHTTP{50701, http.StatusInsufficientStorage, "topic is inactive; no device available to recieve message", ""}
```
karmanyaahm, line 95: I need to add 507 to the UP spec and discuss it with the group further. For now, this can stay.

server/topic.go
```diff
@@ -87,8 +111,9 @@ func (t *topic) CancelSubscribers(exceptUserID string) {
 	t.mu.Lock()
 	defer t.mu.Unlock()
 	for _, s := range t.subscribers {
-		if s.userID != exceptUserID {
-			log.Tag(tagSubscribe).Field("topic", t.ID).Debug("Canceling subscriber %s", s.userID)
+		if s.visitor.MaybeUserID() != exceptUserID {
+			// TODO: Shouldn't this log the IP for anonymous visitors? It was s.userID before my change.
+			log.Tag(tagSubscribe).Field("topic", t.ID).Debug("Canceling subscriber %s", s.visitor.MaybeUserID())
```
karmanyaahm, line 116: Should that log be visitorID instead of userID, to also log anonymous visitors properly?

binwiederhier:
I looked, I promise. I just cannot get my brain into the right headspace, which I need for this. I promise I will :-)

server/errors.go
binwiederhier: "Insuffiencient storage" implies that it's stored on the server, and a "5xx" error implies that it's the server's fault that the request was denied. Also, 507 seems WebDAV specific.

Since there are two clients involved (sender (= traditional client) ---> ntfy server --> subscriber), it's hard to find a good HTTP code, but I'd still say it's a 4xx, no?

Maybe a 409?

> 409 Conflict - Indicates that the request could not be processed because of conflict in the current state of the resource, such as an [edit conflict](https://en.wikipedia.org/wiki/Edit_conflict) between multiple simultaneous updates.

Or a 412? Probably less so..

> 412 Precondition Failed - The server does not meet one of the preconditions that the requester put on the request header fields.

Or 422?

> 422 Unprocessable Entity - The request was well-formed but was unable to be followed due to semantic errors.

I think 409 and 422 work well.

server/server.go
```diff
@@ -114,13 +114,15 @@ var (
...
-	encodingBase64           = "base64"                  // Used mainly for binary UnifiedPush messages
-	jsonBodyBytesLimit       = 16384
+	firebaseControlTopic        = "~control"                // See Android if changed
+	firebasePollTopic           = "~poll"                   // See iOS if changed
+	emptyMessageBody            = "triggered"               // Used if message body is empty
+	newMessageBody              = "New message"             // Used in poll requests as generic message
+	defaultAttachmentMessage    = "You received a file: %s" // Used if message body is empty, and there is an attachment
+	encodingBase64              = "base64"                  // Used mainly for binary UnifiedPush messages
+	jsonBodyBytesLimit          = 16384
+	subscriberBilledTopicPrefix = "up_"
```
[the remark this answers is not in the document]
binwiederhier: 1. We have to find a better word for "billing". It is confusing with all the payment work in the codebase. I do like the term "subscriberBilling", but I think the more semantically correct term would be "subscriberBasedRateLimiting" (albeit a little long)

2. I love your example of how it's useful elsewhere. And it's true, I can see that. Uhhh, what if it's not based on a prefix at all, but on a header that the subscriber sends, like 

```
GET /up123/json HTTP/1.1
Rate-Limits: subscriber
```

server/topic.go
```diff
@@ -32,22 +36,42 @@ func newTopic(id string) *topic {
...
+func (t *topic) getBillee() *visitor {
+	for _, this_subscriber := range t.subscribers {
+		return this_subscriber.visitor
+	}
+	if t.lastVisitor != nil && t.lastVisitorExpires.After(time.Now()) {
+		t.lastVisitor = nil
+	}
+	return t.lastVisitor
+
+}
```
binwiederhier, line 64: 1. Style: Don't use snake_case, use camelCase. And: don't use getters (no getX), use `X` instead, here: `func billee()`
2. This function should use the mutex!
3. This is a strange way of retrieving the billee, IMHO. Why not just keep XXXXXXXXXXXXX aahh toddler. WILL CONTINUE LATER

binwiederhier: Updated my comment from this afternoon.

binwiederhier (changes requested):
1. I think this is a good start, but needs quite some work. I am happy to help and work closely with you. I do want to do this very quickly, because there are very few things I have to do before I can start the payments.

2. This needs UNIT TESTS! A lot of them :-)

server/server.go
```diff
@@ -664,6 +675,8 @@ func (s *Server) handlePublishWithoutResponse(r *http.Request, v *visitor) (*mes
 		if err := s.messageCache.AddMessage(m); err != nil {
 			return nil, err
 		}
+	} else {
+		m.Expires = m.Time
```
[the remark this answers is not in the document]
binwiederhier: Did this in https://github.com/binwiederhier/ntfy/commit/96439ac41f23a793a5c97e73653dcdb1da6dd52b

server/server.go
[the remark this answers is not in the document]
karmanyaahm: 2: Good idea. The ntfy app subscribes to all the topics in one request, so it can't be a global header. Since we only want some topics to be subscriber-based-rate-limited, an alternative could be making a `Subscriber-Rate-Limit-Topics` header?

a. In `parseSubscribeParam` get a list of Subscriber-Rate-Limit-Topics.
b. In topic.Subscribe, pass whether that topic is in the list.
c. The topic can handle getBillee (or whatever that is renamed to accordingly)

Does that sound good?

server/errors.go
karmanyaahm: IMHO by not it's not a 4xx since there's nothing the app server can do about this. It's the push server's "fault" (design limitation) that it doesn't know whether the subscription is valid or not.

Regardless, let's discuss this further in https://github.com/UnifiedPush/specifications/issues/34

binwiederhier:
Cool. I'll start reviewing again, but you gotta repoint to `main` and rebase/merge the latest main.

karmanyaahm:
I'm working on the rest of the things

server/server.go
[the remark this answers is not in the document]
binwiederhier: Yup, that sounds perfect, except for the really long header name. But we can fix that later. I do think that `up*` topics can temporarily be automatically subscriber-rate-limited, so maybe build that in, too?

server/errors.go
karmanyaahm: I'll make it 409 here for now. This should be trivial to update once the spec is ready. Also, really sorry for the delay, I'll try to implement all the comments by noon-ish today.

binwiederhier: No worries dude. It's all my fault. I didn't look at your PR for ages and now I'm rushing you. It's unfair ... I can help with the code and/or tests if you like. Maybe I can write some tests already, TDD style :-)

karmanyaahm: Tests would absolutely help!

server/server.go
```diff
@@ -637,8 +648,9 @@ func (s *Server) handlePublishWithoutResponse(r *http.Request, v *visitor) (*mes
 			"message_email":       email,
 		}).
 		Debug("Received message")
+		//Where should I log the original visitor vs the billing visitor
 	if log.IsTrace() {
-		logvrm(v, r, m).
+		logvrm(v_old, r, m).
```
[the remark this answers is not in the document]
karmanyaahm: Split v and vRate

server/topic.go
karmanyaahm: 1 and 2 are done, and I *think* the latest rewrite achieves 3?

server/util.go
```diff
@@ -35,6 +47,13 @@ func readHeaderParam(r *http.Request, names ...string) string {
 	return ""
 }
 
+func readHeaderParamValues(r *http.Request, names ...string) (values []string) {
+	for _, name := range names {
+		values = append(values, r.Header.Values(name)...)
+	}
+	return
+}
+
```
binwiederhier, line 56: not used

server/server.go
```diff
@@ -114,13 +115,15 @@ var (
...
-	encodingBase64           = "base64"                  // Used mainly for binary UnifiedPush messages
-	jsonBodyBytesLimit       = 16384
+	firebaseControlTopic        = "~control"                // See Android if changed
+	firebasePollTopic           = "~poll"                   // See iOS if changed
+	emptyMessageBody            = "triggered"               // Used if message body is empty
+	newMessageBody              = "New message"             // Used in poll requests as generic message
+	defaultAttachmentMessage    = "You received a file: %s" // Used if message body is empty, and there is an attachment
+	encodingBase64              = "base64"                  // Used mainly for binary UnifiedPush messages
+	jsonBodyBytesLimit          = 16384
+	subscriberBilledTopicPrefix = "up_"
```
binwiederhier, line 125: rename to `unifiedPushTopicPrefix = "up"` (not `up_`!)

server/server.go
```diff
@@ -603,15 +606,23 @@ func (s *Server) handlePublishWithoutResponse(r *http.Request, v *visitor) (*mes
...
+	vRate := v
+	if topicCountsAgainst := t.Billee(); topicCountsAgainst != nil {
+		vRate = topicCountsAgainst
+	}
+
+	if !vRate.MessageAllowed() {
+		vRate = v
+		if !v.MessageAllowed() {
+			return nil, errHTTPTooManyRequestsLimitMessages
+		}
```
binwiederhier, line 618: If `v == vRate`, then this calls `v.MessageAllowed()` twice, no?

karmanyaahm: Don't need that anyway if we're assuming a secret topic, and the subscriber taking the full load, so I got rid of the two stage process.

server/util.go
karmanyaahm: mb

server/server.go
```diff
@@ -603,6 +605,14 @@ func (s *Server) handlePublishWithoutResponse(r *http.Request, v *visitor) (*mes
 	if err != nil {
 		return nil, err
 	}
+	v_old := v
+	if strings.HasPrefix(t.ID, subscriberBilledTopicPrefix) {
+		v = t.getBillee()
+		if v == nil {
+			return nil, errHTTPWontStoreMessage
+		}
+	}
```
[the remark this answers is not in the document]
karmanyaahm: I did this using context (as suggested below), but isn't that supposed to be an anti-pattern? In any case, I guess it's just one thing so it shouldn't matter much. I feel like the ideal solution to this would be making `visitor` an interface and having a `compositeVisitor` in addition to visitor...but, that's probably overengineering.

binwiederhier:
Superseded by https://github.com/binwiederhier/ntfy/pull/633

PR #623, opened 2023-02-20
allow publishing of text/html emails

tamcore:
This allows publishing of emails with `Content-Type: text/html` type, by quickly converting them to plaintext. I ran into this issue, when i tried to point my [traccar](https://github.com/traccar/traccar) instance's email notifications towards my ntfy instance.

Container image to test `ghcr.io/tamcore/ntfy@sha256:caa2c16b9eef9fd6a3d32e9b4621bb32218198fb5931481f3b96ce72819f22b0`

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -27,6 +27,7 @@ require github.com/pkg/errors v0.9.1 // indirect
 
 require (
 	firebase.google.com/go/v4 v4.10.0
+	github.com/k3a/html2text v1.1.0
 	github.com/stripe/stripe-go/v74 v74.8.0
 )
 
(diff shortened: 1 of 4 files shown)


binwiederhier:
While that would be a nice feature to have, I am not willing to pull in a dependency for this, especially not one that only implements 300 lines of code and has 78 stars. But don't bother searching for a more prestigious project because I really don't think bringing in a lib for this is is something I want to do. 

Aside from the library issue, there is another one: this now prefers html content if it comes first, which it usually does. So you'd have to do something with ordering. 

All in all, unless you find a stdlib way to so this, i don't think I am willing to accept this PR. 

Feel free to create a ticket, and if a few people like it, i may reconsider.

tamcore:
Understood. I guess a bit of basic regex'ing to strip the HTML code could be an option? Could also move the `text/html` condition bellow `multipart/`, which then should fix the ordering issue.

binwiederhier:
**Re regex:** I am not sure how I feel about the regex-based HTML stripping. I suppose since we're not ever evaluating the HTML it's fine, but I don't know what cases it covers. I'm more of a "do it right, or don't do it at all" kinda guy, so not sure.

I think if you can prove to me with some real HTML emails (check your inbox) that your logic will produce good-enough results (in a unit test), I'm gonna accept the PR. Otherwise I think I'd rather not have it. Nobody has asked for this before, so I'm hesitant.

**Re ordering:** I don't think the re-ordering of the if-statements will achieve the text/plain-affinity that I'd like. If we have a multipart message like this:

```
<multipart header>
<text/html part>
<text/plain part>
```

Then we'll still pick the text/html part first.

PR #631, opened 2023-02-22
docs: add traccar example

tamcore:
This will add a configuration example to integrate traccar with ntfy. 

Based on https://github.com/traccar/traccar/pull/4921#issuecomment-1439759339 and https://www.traccar.org/forums/topic/ntfysh/ I guess there's a bit of an interest there :)


binwiederhier:
Cool. Thanks.

PR #633, opened 2023-02-23
WIP: Subscriber rate limiting

binwiederhier:


server/server.go
```diff
@@ -403,13 +405,14 @@ func (s *Server) handleInternal(w http.ResponseWriter, r *http.Request, v *visit
 	} else if r.Method == http.MethodOptions {
 		return s.limitRequests(s.handleOptions)(w, r, v) // Should work even if the web app is not enabled, see #598
 	} else if (r.Method == http.MethodPut || r.Method == http.MethodPost) && r.URL.Path == "/" {
-		return s.limitRequests(s.transformBodyJSON(s.authorizeTopicWrite(s.handlePublish)))(w, r, v)
+		// So I don't *really* have to switch this order, since this is unrelated to UP; But, making this and matrix inconsistent is just calling for confusion, no?
+		return s.transformBodyJSON(s.limitRequestsWithTopic(s.authorizeTopicWrite(s.handlePublish)))(w, r, v)
 	} else if r.Method == http.MethodPost && r.URL.Path == matrixPushPath {
-		return s.limitRequests(s.transformMatrixJSON(s.authorizeTopicWrite(s.handlePublishMatrix)))(w, r, v)
+		return s.transformMatrixJSON(s.limitRequestsWithTopic(s.authorizeTopicWrite(s.handlePublishMatrix)))(w, r, v)
```
binwiederhier, line 411: This is actually causing an issue: We are now not returning any non-200 HTTP codes anymore. I get why this has to be this way. Gotta just find a solution for this.

binwiederhier:
@karmanyaahm I cleaned things up a little and added tests. There are still some issues, but I think this works nicely.

server/topic.go
```diff
@@ -32,18 +40,43 @@ func newTopic(id string) *topic {
...
+
+func (t *topic) RateVisitor() *visitor {
+	t.mu.Lock()
+	defer t.mu.Unlock()
+	if t.rateVisitorExpires.Before(time.Now()) {
+		t.rateVisitor = nil
+	}
+	return t.rateVisitor
+}
+
```
karmanyaahm, line 79: What happens if someone is continuously subscribed for longer than 12 hours (or `rateVisitorExpiryDuration`)? Shouldn't  the countdown only start after rateVisitor unsubscribed?

Also, I removed the expiry check in RateVisitor (ex. Billee) because it runs minutely through the manager/Stale anyway. The comparison is probably low cost though so it doesn't matter.

binwiederhier: > Shouldn't the countdown only start after rateVisitor unsubscribed?

I see. What about this:

- The `topic.expires` stays 0 until somebody unsubscribes
- `topic.Stale()` returns true if `len(subs) == 0 && expires has passed`

So we don't tie the topic expiry to the rate visitor at all anymore. We just make sure the topic sticks around for a few hours, even after everybody disconnects. 

I shortened the expiry time to 6 hours, which should be enough buffer time anyway. Realistically, the topic won't get cleaned up anyway, as long as messages are in the topic (see manager).

https://github.com/binwiederhier/ntfy/pull/633/commits/8eae44ea614483c912f6484cfb2295d12096fec1

PR #637, opened 2023-02-25
Subscriber Rate Limiting Error Handling

karmanyaahm:
As discussed in #unifiedpush-devs:matrix.org

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -661,7 +661,7 @@ func (s *Server) handlePublish(w http.ResponseWriter, r *http.Request, v *visito
 func (s *Server) handlePublishMatrix(w http.ResponseWriter, r *http.Request, v *visitor) error {
 	_, err := s.handlePublishWithoutResponse(r, v)
 	if err != nil {
-		return &errMatrix{pushKey: r.Header.Get(matrixPushKeyHeader), err: err}
+		return err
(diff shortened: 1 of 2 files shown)


server/server.go
```diff
@@ -661,7 +661,7 @@ func (s *Server) handlePublish(w http.ResponseWriter, r *http.Request, v *visito
 func (s *Server) handlePublishMatrix(w http.ResponseWriter, r *http.Request, v *visitor) error {
 	_, err := s.handlePublishWithoutResponse(r, v)
 	if err != nil {
-		return &errMatrix{pushKey: r.Header.Get(matrixPushKeyHeader), err: err}
+		return err
```
karmanyaahm, line 664: Already wrapped in `transformMatrixJSON`, so no need to double wrap.

karmanyaahm (commented):
1. Fixed Matrix error wrapping and pushkeys
2. Subscribe to topic before running UP tests
3. Add a Matrix test for 507 case
4. Change a Matrix test to avoid rejecting for rate limits

karmanyaahm:
Also, if you really don't like 507, any other 5xx code should do. We can even make up our own new code.

binwiederhier:
Thanks

Issue #650, opened 2023-03-03
Token authentication doesn't work if default user credentials are defined in client.yml


Xinayder:
:lady_beetle: **Describe the bug**

Trying to use ntfy CLI with token authentication seems to fail if the `client.yml` settings file has default user credentials configured, i.e., the fields `default-user` and `default-password` are set.

:computer: **Components impacted**

ntfy server

:bulb: **Screenshots and/or logs**

Setting `client.yml` with `default-user: user` and `default-password: password`:

```
$ ntfy pub -d --token=tk_cthk85i38xaP03X1rmhXEiM91xLLc https://notify.mydomain.com/teste testing message

2023/03/03 16:30:53 DEBUG notify.mydomain.com/teste Publishing message with headers map[Authorization:[Basic dXNlcjpwYXNzd29yZAo=]]
{"code":40101,"http":401,"error":"unauthorized","link":"https://ntfy.sh/docs/publish/#authentication"}
```

:crystal_ball: **Additional context**

```
$ ntfy token list

user user
- tk_cthk85i38xaP03X1rmhXEiM91xLLc, never expires, accessed from 0.0.0.0 at 03 Mar 23 16:29 UTC
```

PR #651, opened 2023-03-03
Fix publish command preferring default user instead of token auth

Xinayder:
As reported in #650, the `publish` command seems to prefer `default-user` and `default-password` config keys if they are defined in `client.yml`. 

This PR aims to fix this problem, checking if the token isn't passed as a flag to the `ntfy pub` command before attempting to use the credentials stored in the client config file.


binwiederhier:
Very nice, and very simple. Thank you.

Issue #649, opened 2023-03-03
Error handling Matrix request internal server error: cannot publish to UnifiedPush topic without previously active subscriber


barathrm:
:lady_beetle: **Describe the bug**

Matrix notifications no longer work. Other notifications work fine. Way I tested matrix notifications

- make sure matrix apps are using ntfy as the notification backend
- make sure other ntfy notifications work (using non-UP channels)
- send a message from another account to my main account, which schildichat is logged in with
- observe ntfy server logs
- observe no notification for schildichat popping up

:computer: **Components impacted**

- matrix.org matrix account
- self-hosted ntfy server 
- android phone running schildichat, fluffychat, android ntfy app.

:bulb: **Screenshots and/or logs**

> DEBUG Error handling Matrix request (error=internal server error: cannot publish to UnifiedPush topic without previously active subscriber, error_code=50701, http_method=POST, http_path=/_matrix/push/v1/notify, http_status=507, tag=matrix, topic=<snip>, topic_subscribers=1, visitor_auth_limiter_limit=0.016666666666666666, visitor_auth_limiter_tokens=10, visitor_emails=0, visitor_emails_limit=24, visitor_emails_remaining=24, visitor_id=ip:<snip>, visitor_ip=<snip>, visitor_messages=0, visitor_messages_limit=17280, visitor_messages_remaining=17280, visitor_request_limiter_limit=0.2, visitor_request_limiter_tokens=59.0001245076)

:crystal_ball: **Additional context**

The problem only occurs on ntfy `2.1.1` and `2.1.0`. **Rolling back to `1.31.0` fixes the issue**.

Let me know if there's anything I can do to help debug.

binwiederhier:
Thanks for reporting this. While I do think this may actually be a bug, the way you describe it sounds like "intended behavior", so let me clarify:

Did you subscribe to the topic before sending the message? If you have debugging enabled, you should see some `GET /up1234/(json|ws)` message and something that says "registering rate visitor". Maybe post a more extensive excerpt from the logs?

@S1m @karmanyaahm I do think we'll have to urgently change this logic again if it is actually causing delivery issues. Maybe the 400k Matrix messages are not all dead letters after all.

barathrm:
> Did you subscribe to the topic before sending the message?

Well, I'm not sure. I never completely understood how subscribing to a topic works wrt. UP and matrix clients. 

[...]

PR #654, opened 2023-03-03
WIP: Enable subscriber rate limiting

binwiederhier:


binwiederhier:
Implements what was discussed in #649

PR #659, opened 2023-03-06
allow default-token and per-subscription tokens in client.yml

### LINKED ISSUE #653: Add `default-token` option to `client.yml`
It would be nice if we had the option to use a default token in `client.yml` instead of default user/default password

Adding token options for specific subscriptions would be nice, too

wunter8:
fixes #653

### DIFF
--- a/cmd/publish_test.go
+++ b/cmd/publish_test.go
@@ -5,8 +5,11 @@ import (
 	"github.com/stretchr/testify/require"
 	"heckel.io/ntfy/test"
 	"heckel.io/ntfy/util"
+	"net/http"
+	"net/http/httptest"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"strconv"
 	"strings"
 	"testing"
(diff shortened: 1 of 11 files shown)


cmd/publish_test.go
```diff
@@ -130,7 +133,7 @@ func TestCLI_Publish_Wait_PID_And_Cmd(t *testing.T) {
 	require.Equal(t, `command failed: does-not-exist-no-really "really though", error: exec: "does-not-exist-no-really": executable file not found in $PATH`, err.Error())
 
 	// Tests with NTFY_TOPIC set ////
-	require.Nil(t, os.Setenv("NTFY_TOPIC", topic))
```
wunter8, line 133: The tests after this one were failing because they kept connecting to the wrong ntfy server. It took me hours to figure out that this was the reason; the environment variable was persisting after this test and was affecting the remaining tests in the file.

wunter8:
I realize it's a bit sloppy to include the latest commit here instead of in a second PR, but I needed it to pass the tests and eliminate the race condition in the old version of the tests here.

cmd/publish_test.go
binwiederhier: I didn't know you can do that `t.Setenv`. Wow. Sorry about that.

binwiederhier (approved):

binwiederhier:
I just merged this 600 line PR without a single comment other than "wow i didn't know that". Those that know me know that that is pretty much an impossible thing to achieve. Great work and 100 thanks.

wunter8:
I didn't even realize this was 600 lines, haha. I knew it was big, but I didn't expect it was _that_ big

PR #661, opened 2023-03-07
convert query parameters to lowercase before reading them

### LINKED ISSUE #630: Query parameters accept only lowercase parameter names
The parameter names are described as case-insensitive in the documentation, but only lowercase is accepted when passed by the query parameters

wunter8:
fixes #630

### DIFF
--- a/server/server_test.go
+++ b/server/server_test.go
@@ -309,6 +309,18 @@ func TestServer_PublishGETOnlyOneTopic(t *testing.T) {
 	require.Equal(t, 404, response.Code)
 }
 
+func TestServer_PublishCaseInsensitiveQueryParam(t *testing.T) {
+	s := newTestServer(t, newTestConfig(t))
+	response := request(t, s, "GET", "/mytopic/publish?MESSAGE=testing", "", nil)
+	require.Equal(t, "testing", toMessage(t, response.Body.String()).Message)
+}
+
+func TestServer_PublishInvalidQueryParam(t *testing.T) {
+	s := newTestServer(t, newTestConfig(t))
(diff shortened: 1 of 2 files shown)


wunter8:
I forgot to modify the docs, and since I pushed this from main, it doesn't have the doc changes from my other PRs, and I don't want to worry about it trying to auto-merge, haha. If you want to add a line to the docs for this, here it is (but I'm already a little self-conscious about how many times my name is included in this release, so it's fine if you leave it out):

```
* Query parameters when publishing via a GET request (e.g., `/mytopic/publish?Message=testing`) are now case-insensitive ([#630](https://github.com/binwiederhier/ntfy/issues/630) thanks to [@bbaa-bbaa](https://github.com/bbaa-bbaa) for reporting, and [@wunter8](https://github.com/wunter8) for fixing)
```

binwiederhier:
I kinda hate this "fix" for two reasons:

1. It's a pretty unnecessary feature to have. We could have just adjusted to docs to say "look, you gotta have it all lowercase for query params"
2. It's not efficient, since it loops through every param every time readParam is called. It's probably very much negligible, but I still hate it.

wunter8:
I was concerned about efficiency and did the lowercase-ing outside of the `names` loop, but I didn't consider all the different parameters we check for. That's true.

I think just updating the docs could be a valid change. Most people use query parameters for publishing instead of headers or JSON body because they're using a webhook and the only thing they can control is the URL, right? Which means, they have control over the query parameters and can make them lowercase

binwiederhier:
Let's just do a doc update on this. We can always bring the PR back to life when other people ask for it.

PR #665, opened 2023-03-08
Allow newer python when building docs

cyqsimon:
Previously in `Makefile`, there are provisions to use `python3.8` if `python3` is not new enough. However it doesn't check for versions newer than 3.8, which could be problematic if the user has `python3.6` and `python3.9`, but `python3` points to `python3.6`. This is a frequent occurrence on RHEL systems.

This change also makes sure that the version of `pip` called matches the version of `python` used. Now I'm not that familiar with Python, so I cannot say for certain whether this is strictly necessary. But I cannot imagine there to be any downsides with doing this.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -96,25 +96,28 @@ build-deps-ubuntu:
 docs: docs-deps docs-build
 
 docs-build: .PHONY
-	@if ! /bin/echo -e "import sys\nif sys.version_info < (3,8):\n exit(1)" | python3; then \
-	  if which python3.8; then \
-	  	echo "python3.8 $(shell which mkdocs) build"; \
-	    python3.8 $(shell which mkdocs) build; \
-	  else \
-	    echo "ERROR: Python version too low. mkdocs-material needs >= 3.8"; \
-	    exit 1; \
-	  fi; \
-	else \
-	  echo "mkdocs build"; \
-	  mkdocs build; \
-	fi
+	@if ! SUFFIX=$$(tools/get-best-python-for-docs-build.sh); then \
+		echo "ERROR: could not find a recent version of Python. mkdocs needs >= 3.8"; \
+		exit 1; \
+	fi; \
+	echo "python$$SUFFIX $$(which mkdocs) build"; \
+	"python$$SUFFIX" $$(which mkdocs) build
 
 docs-deps: .PHONY
-	pip3 install -r requirements.txt
+	@if ! SUFFIX=$$(tools/get-best-python-for-docs-build.sh); then \
+		echo "ERROR: could not find a recent version of Python. mkdocs needs >= 3.8"; \
+		exit 1; \
+	fi; \
+	echo "pip$$SUFFIX install -r requirements.txt"; \
+	"pip$$SUFFIX" install -r requirements.txt
 
 docs-deps-update: .PHONY
-	pip3 install -r requirements.txt --upgrade
-
+	@if ! SUFFIX=$$(tools/get-best-python-for-docs-build.sh); then \
+		echo "ERROR: could not find a recent version of Python. mkdocs needs >= 3.8"; \
+		exit 1; \
+	fi; \
+	echo "pip$$SUFFIX install -r requirements.txt --upgrade"; \
+	"pip$$SUFFIX" install -r requirements.txt --upgrade
 
 # Web app
 
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,10 @@
 MAKEFLAGS := --jobs=1
+
 VERSION := $(shell git describe --tag)
 COMMIT := $(shell git rev-parse --short HEAD)
+
+PY_BIN := $(shell tools/get-python-bin.sh python)
+PIP_BIN := $(shell tools/get-python-bin.sh pip)
 
 .PHONY:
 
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -96,25 +96,28 @@ build-deps-ubuntu:
...
 
 docs-deps-update: .PHONY
-	pip3 install -r requirements.txt --upgrade
-
+	@if ! SUFFIX=$$(tools/get-best-python-for-docs-build.sh); then \
+		echo "ERROR: could not find a recent version of Python. mkdocs needs >= 3.8"; \
+		exit 1; \
+	fi; \
+	echo "pip$$SUFFIX install -r requirements.txt --upgrade"; \
+	"pip$$SUFFIX" install -r requirements.txt --upgrade
```
cyqsimon, line 120: If there's a good way to get rid of this duplication I'd love to know.

cyqsimon:
I don't think the test failure has anything to do with this patch...?

binwiederhier:
This is such a terrible hack. Honestly though, the whole python version thing is such a hack. Do you think there is a better solution somehow?

If this is truly the best way, then I think the script should instead return the path to the python/pip binary to use, and not a suffix, e.g.

```
$ tools/python-path.py python
/usr/bin/python3.8

$ tools/python-path.py pip
/usr/bin/pip3.8

# If it fails:
$ tools/python-path.py pip
<empty, exits 1>
```

So you can use it like this in the Makefile:

```
$(tools/python-path.py pip) install -r requirements.txt ...
```

That should fail if the subshell fails, right? I never know with make...

cyqsimon:
> Honestly though, the whole python version thing is such a hack. Do you think there is a better solution somehow?

Yeah I know. I dislike it too. But it's not like Python has an official version manager like Rust's `rustup` =(.

> I think the script should instead return the path to the python/pip binary to use

I agree. I will make changes.

> That should fail if the subshell fails, right? I never know with make...

Don't worry. Everyone hates make 😅.

cyqsimon:
Unfortunately your suspicion that the subshell failure may not propagate is true =(. And there's no easy way to configure this behaviour with Make's `$(shell ...)` function.

This is the cleanest solution I ended up with using the shell's command substitution. I had to assign the script's output to a variable first because that's the only way to make the subshell's error propagate. See https://unix.stackexchange.com/a/613021/375550.

cyqsimon:
Is there anything else that you are not happy with? @binwiederhier 

If it's just that you dislike the "hackiness" of this patch, I can ask around in a few coding communities and see what they can come up with.

tools/get-python-bin.sh
```diff
@@ -0,0 +1,45 @@
...
+
+# if `python3` is >= 3.8 and `pip3` is available, use that
+if echo -e "import sys\nif sys.version_info < (3,8):\n exit(1)" | python3 && \
+which pip3 1>/dev/null 2>&1; then
+    echo "${BIN_PREFIX}3"
+    exit 0
+fi
+
+# check `python3.N` from newest to oldest
+CANDIDATE_SUFFIXES=("3.11" "3.10" "3.9" "3.8")
```
mnuzzose, line 33: This list would have a maintenance overhead from having to update to accept new python versions every time a new version of Python 3 is released. Instead, you should probably do something like `python3 --version | grep "^Python 3\.[0-7][$|\.]"` to check if the version of Python matches a known incompatible version and perform the error actions if it does.

cyqsimon: I understand what you're talking about - in fact I initially wanted to do something like this. This is what I would write if we are to go down this route:

```bash
# compgen is bash-specific, but we asked for bash in shebang so it's fine
CANDIDATE_MINOR_SUFFIXES=$(compgen -c | grep -P '^python3\.[0-9]+$' | sed 's/python3\.//' | awk 'int($NF) >= 8' | sort -nr)
```

`compgen` gives all executables in `$PATH`; `grep` filters for `python3.*`; `sed` strips prefix; `awk` sets minimum version; `sort` orders the versions.

---

The reason I eventually decided against this is for the simple reason that Python is not exactly known for strong future compatibility guarantees. While unlikely, there is a non-insignificant chance that Python 3.14 or something comes along and breaks standard library (for an example of this see https://github.com/bastibe/SoundCard/issues/140).

So in the end I simply verified that all of 3.8, 3.9, 3.10, and 3.11 work, then hard-coded them here. It's just not worth it to try to anticipate what will happen in the future. When 3.12 / 3.13 / 3.14 eventually comes along, it would just be a trivial edit anyways.

cyqsimon: On second thought, maybe the best way to handle this is to automatically use the newest (and possibly future) versions, but also allow overriding in Makefile. This way in case a future python version breaks something, packagers can always set a temporary override without having to wait for an upstream fix.

I'll go ahead and make these changes for now.

binwiederhier:
I removed the if entirely. Time has moved enough for this issue to not matter anymore (hopefully).

cyqsimon:
Hmm. I understand and agree with the decision but I do believe that the original problem still exists, namely that it's not possible to use a newer version of python when `python3` points to an older version.

Rethinking this however, the problem really is quite simple to fix. Just make the binary name a Make variable and allow the user to override it if need be. I'll put in another PR for it.

PR #667, opened 2023-03-12
Fix remove external google font server dependency

ozskywalker:
For reference:
* Once you set `font: false` in mkdocs, it'll fall back on system fonts and stop loading remote
* `extra.css` then is modified to recognize the new fonts (`@font-face`) AND override the default text/code font (`--md-text-font` / `--md-code-font`) to use our new fonts specified by `@font-face` declaration
* Then the final piece is a quick 'n dirty mkdocs plugin to copy the font files into place during static site generation


ozskywalker:
Modified PR to drop support for older browsers (keep WOFF/WOFF2, remove SVG, etc.) based on discord conversation

PR #678, opened 2023-03-21
I18n plural forms

bmoczulski:
While working on Polish translations I noticed that some labels can't be translated correctly due lack of plural forms. In Polish there are two plural forms of nouns - one for cadrinality 2-4 mod 10 (except 12-14) and the another for everything else > 1. Other languages have different rules for plurals, sometimes even more complex. Considering that both react-i18next and Weblate fully support translation of plural forms I propose to change the affected entries by adding `{count: n}` to relevant `t()` calls and `_one/_other` suffix for English translation. This will make both systems aware of plurality of these labels. Weblate will show a form with all required plural variants in a given language and react-i18next will look for appropriate suffixes based on `count` and language.

This is my firs PR here so let me take the opportunity to congratulate you on creating this awesome utility. I'm currently evaluating ntfy in my monitoring deployment and I have to admit I'm impressed by the plethora of features and comprehensive documentation. Well done! The very least I can do in return is to provide some translations in my mother tongue.


binwiederhier:
Thank you. Fixed the other languages too.

PR #697, opened 2023-04-05
Converted PowerShell code to use Splatting, and newer PS7 parameters (where available)

Natfan:
Also removed the requirement for UseBasicParsing, as Invoke-RestMethod already converts XML/JSON to PSObjects as required.

More info on splatting: https://learn.microsoft.com/powershell/module/microsoft.powershell.core/about/about_splatting
More info on Invoke-RestMethod: https://learn.microsoft.com/powershell/module/microsoft.powershell.utility/invoke-restmethod

### DIFF
--- a/docs/publish.md
+++ b/docs/publish.md
@@ -38,7 +38,12 @@ Here's an example showing how to publish a simple message using a POST request:
 
 === "PowerShell"
     ``` powershell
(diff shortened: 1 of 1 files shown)


binwiederhier:
This is great, thank you. I'll likely merge this tomorrow. I gotta say I mostly was winging the PS examples, so I'm glad somebody actually fixed them up :-D

PR #704, opened 2023-04-18
Added support for armel (ARMv5) architecture

ant9000:
Straightforward patch for armel arch (ARMv5 soft-float) support.

Thanks a bunch for your wonderful project!

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -14,6 +14,18 @@ builds:
     goos: [linux]
     goarch: [amd64]
   -
+    id: ntfy_linux_armv5
+    binary: ntfy
+    env:
+      - CGO_ENABLED=1 # required for go-sqlite3
+      - CC=arm-linux-gnueabi-gcc # apt install gcc-arm-linux-gnueabi
+    tags: [sqlite_omit_load_extension,osusergo,netgo]
+    ldflags:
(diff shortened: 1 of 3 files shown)


ant9000:
Wasn't able to find any reference to ARMv5/armel in past issues, but I see your point:  Following your suggestion

ant9000:
Wasn't able to find any reference to ARMv5/armel in past issues, but I see your point with regards to aptly: you have two different packages for the same architecture, since ARMv6 and ARMv7 are both armhf.
This should not be a problem with ARMv5, since the architecture is armel, which is indeed different.
I've done a `make release-snapshot`  and spotted a few places where my first patch was incomplete - now it generates all the packages, and not only the binary executable.

.goreleaser.yml
```diff
@@ -190,16 +203,26 @@ dockers:
     goarm: 6
     build_flag_templates:
       - "--platform=linux/arm/v6"
+  - image_templates:
+      - &armv5_image "binwiederhier/ntfy:{{ .Tag }}-armv5"
+    use: buildx
+    dockerfile: Dockerfile
+    goarch: arm
+    goarm: 6
```
binwiederhier, line 211: 6?

ant9000: Whoopsie :blush:  Thanks!

binwiederhier:
Nobody has asked in 3 years, so I am closing this. Happy to address if there is interest in the future.

Issue #712, opened 2023-05-01
server/topic.Subscribe can produce duplicate subscriber IDs


peterbourgon:
https://github.com/binwiederhier/ntfy/blob/6ad3b2e8020082a5b7206d36d425f925354cd059/server/topic.go#L50

```go
// Subscribe subscribes to this topic
func (t *topic) Subscribe(s subscriber, userID string, cancel func()) int {
	t.mu.Lock()
	defer t.mu.Unlock()
	subscriberID := rand.Int() // <---- here
	t.subscribers[subscriberID] = &topicSubscriber{
		userID:     userID, // May be empty
		subscriber: s,
		cancel:     cancel,
	}
	t.lastAccess = time.Now()
	return subscriberID
}
```

Nothing prevents `subscriberID := rand.Int()` from producing an integer which is already present in the `t.subscribers` map. In that case, the previous subscriber is over-written.

dropdevrahul:
can I pick this up?

dropdevrahul:
@binwiederhier tried a fix with https://github.com/binwiederhier/ntfy/pull/713

dropdevrahul:
Please not Just added a simple retry mechanism and added a test case to catch the edge case

binwiederhier:
Thanks to you both. The best effort retry should be enough.

PR #713, opened 2023-05-02
fix: removes an issue with topic.Subscribe function not checking dupl…

dropdevrahul:
…icate ID.

Fixes issue https://github.com/binwiederhier/ntfy/issues/712


binwiederhier:
Thanks

PR #719, opened 2023-05-08
Add quoted-printable decoding to smtp server

Aerion:
Some e-mails are sent using quoted-printable encoding [0], resulting in notifications with weird characters.

This commit adds support for this encoding, resulting in the following:

**Before**
```
A
=3D=3D=3D=3D=3D
B
=3D=3D=3D=3D=3D
C
```

**After**
```
A
=====
B
=====
C
```

[0] https://www.rfc-editor.org/rfc/rfc2045.html

### DIFF
--- a/server/smtp_server.go
+++ b/server/smtp_server.go
@@ -9,6 +9,7 @@ import (
 	"io"
 	"mime"
 	"mime/multipart"
+	"mime/quotedprintable"
 	"net"
 	"net/http"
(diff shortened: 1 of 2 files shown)


binwiederhier:
That's a good catch. I was under the impression that quoted-printable is handled automatically, but apparently that only happens for multipart parts (see https://pkg.go.dev/mime/multipart#Reader.NextPart).

PR #721, opened 2023-05-12
Fix false positive in ContainsAll function

adamantike:
As the `ContainsAll` is working with a match counter, it could return a false positive when the `haystack` slice contains duplicate elements.

This can be checked with the included testing scenario, with `haystack = [1, 1]` and `needles = [1, 2]`. Iterating over the haystack to check for items to be present in needles will increase the match counter to 2, even if `2` is not present in the first slice.

### DIFF
--- a/util/util.go
+++ b/util/util.go
@@ -6,7 +6,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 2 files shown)


binwiederhier:
Thanks :-D

PR #725, opened 2023-05-13
Convert mailer_emoji JSON file to map

adamantike:
This fixes a pending TODO comment regarding inefficient tags to emojis mapping, by requiring a full scan over emoji aliases to determine matches.

Instead, now the JSON file is a map, with aliases as keys, and emojis as values. The script to convert the file with Python was:

```python
import json

with open("./server/mailer_emoji.json", "r", encoding="utf-8") as f:
    content = json.load(f)

emoji_map = {}
for emoji in content:
    for alias in emoji["aliases"]:
        if alias in emoji_map:
            print("WARNING: Duplicate alias:", alias)
            continue
        emoji_map[alias] = emoji["emoji"]

sorted_emoji_map = {k: emoji_map[k] for k in sorted(emoji_map)}

with open("./server/mailer_emoji_map.json", "w", encoding="utf-8") as f:
    json.dump(sorted_emoji_map, f, indent=4, ensure_ascii=False)
```


binwiederhier:
Thanks

PR #728, opened 2023-05-14
set attachment filename when download through browser

### LINKED ISSUE #726: Port missing in URL for "send notification" via web gui / attachement
Port missing in URL for "send notification" via web gui / attachement see screenshot:

grafik

grafik

should look like https://ntfy.example.org:9999/file/Xx7CQB4e8c25.exe, right?

BTW: is it intended that file name in link is not reflecting original filename?

wunter8:
Fixes #726

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -642,6 +642,9 @@ func (s *Server) handleFile(w http.ResponseWriter, r *http.Request, v *visitor) 
 		return err
 	}
 	defer f.Close()
+	if m.Attachment.Name != "" {
+		w.Header().Set("Content-Disposition", "attachment; filename="+strconv.Quote(m.Attachment.Name))
+	}
 	_, err = io.Copy(util.NewContentTypeWriter(w, r.URL.Path), f)
 	return err
 }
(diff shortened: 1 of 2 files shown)


server/server.go
```diff
@@ -642,6 +642,9 @@ func (s *Server) handleFile(w http.ResponseWriter, r *http.Request, v *visitor)
 		return err
 	}
 	defer f.Close()
+	if m.Attachment.Name != "" {
+		w.Header().Set("Content-Disposition", "attachment; filename="+strconv.Quote(m.Attachment.Name))
+	}
```
binwiederhier, line 647: Wait, ... This changes the default browser behavior, right? Will images be downloaded too, or just things that are normally downloaded?

wunter8: Everything will download automatically if you go directly to the URL, including images. When would you go to the URL and not want to download it? Just to see a bigger version of the image?

We could add a condition based on content type

binwiederhier: Well, the web app embeds images. I am not sure if there are other use cases, but I'd say it's safe to say we'll want to inline display images at least. Hm....

wunter8: I'm 90% sure the embedded/inline images worked just fine in the web app when I tested it

binwiederhier: Oh interesting. Maybe the browser does the right thing. I'll test.

binwiederhier: Works indeed.

PR #734, opened 2023-05-18
Add native arrsuite & shoutrrr docs

nimbleghost:
Radarr, Sonarr v4, and Prowlarr no longer _require_ the use of custom shell scripts as they have native support.


binwiederhier:
Thanks!

PR #743, opened 2023-05-23
[web] remove unused @emotion packages

nimbleghost:
I was just looking around and realized these were unused.


binwiederhier:
Thanks

PR #746, opened 2023-05-23
Add prettier

nimbleghost:
Adds prettier to npm, plus the relevant make targets.

Created a `devDependencies` sections for tooling, and moved react-scripts there too, which is the only other non-production dependency.

This PR needs to be rebased or merged with a merge commit into master, so that the `.git-blame-ignore-refs` file still references a valid commit.


nimbleghost:
Hmm, depending on what weblate does, perhaps we want to ignore the translations in prettier too.

binwiederhier:
@nimbleghost Just checking, you didn't change anything but the formatting, right? I'm going through the PR but it doesn't look like it.

nimbleghost:
@binwiederhier Correct, just `npm run format` in this PR!

binwiederhier:
It's already merged. I do have to check everything even if it's autogenerated. Due diligence.

PR #748, opened 2023-05-24
Add eslint with eslint-config-airbnb

nimbleghost:
(Please look at the individual commit descriptions for more context)


web/src/app/AccountApi.js
```diff
@@ -261,25 +261,25 @@ class AccountApi {
...
     console.log(`[AccountApi] Creating billing subscription with ${tier} and interval ${interval}`);
-    return await this.upsertBillingSubscription("POST", tier, interval);
+    return this.upsertBillingSubscription("POST", tier, interval);
   }
 
   async updateBillingSubscription(tier, interval) {
     console.log(`[AccountApi] Updating billing subscription with ${tier} and interval ${interval}`);
-    return await this.upsertBillingSubscription("PUT", tier, interval);
+    return this.upsertBillingSubscription("PUT", tier, interval);
   }
```
nimbleghost, line 270: (`return await` is equivalent to async function `return` when there is no `catch`)

web/src/app/AccountApi.js
```diff
@@ -322,12 +322,12 @@ class AccountApi {
       headers: withBearerAuth({}, session.token()),
       body: JSON.stringify({
         number: phoneNumber,
-        code: code,
+        code,
       }),
     });
   }
 
-  async deletePhoneNumber(phoneNumber, code) {
```
nimbleghost, line 330: unused, not specified at call site either.

web/src/app/Connection.js
```diff
@@ -1,7 +1,14 @@
...
 import { basicAuth, bearerAuth, encodeBase64Url, topicShortUrl, topicUrlWs } from "./utils";
 
 const retryBackoffSeconds = [5, 10, 20, 30, 60, 120];
 
+export class ConnectionState {
+  static Connected = "connected";
+
+  static Connecting = "connecting";
+}
+
```
binwiederhier, line 11: Can you explain why you move this up?

web/src/app/Connection.js
```diff
@@ -63,7 +70,7 @@ class Connection {
         this.ws = null;
       } else {
         const retrySeconds = retryBackoffSeconds[Math.min(this.retryCount, retryBackoffSeconds.length - 1)];
-        this.retryCount++;
+        this.retryCount += 1;
```
binwiederhier, line 73: Can you elaborate on why this is better? i++ seems clearer to me and verrrry common.

web/src/app/AccountApi.js
binwiederhier: I asked ChatGPT and I still don't fully understand, but this seems to be legit.

> In JavaScript, the await keyword is used to wait for a promise to resolve and return its resolved value. However, if a function or method is not using the await keyword, it means that it does not need to wait for the promise to resolve and can continue execution without blocking. Therefore, removing the await keyword in this case does not change the behavior of the code, and the change is considered equivalent.

Before I read this, I would have thought that "my function" returns the actual thing, whereas your function returns a promise. No?

binwiederhier: ChatGPT cont'd:

> The original version of the code uses the await keyword before this.upsertBillingSubscription("PUT", tier, interval), which indicates that the updateBillingSubscription method is an asynchronous function. It waits for the promise returned by upsertBillingSubscription to resolve before returning the resolved value.

> The modified version removes the await keyword, making it a regular function that returns a Promise. The Promise will resolve to the value returned by upsertBillingSubscription.

> So, the two versions are not completely equivalent in terms of the return value. The version with await returns the resolved value directly, while the version without await returns a Promise that resolves to the value.

Sooooo, your version returns the Promise, while mine returns the thing. But that doesn't matter because it has to be called with `await` anyway? Is that right?

web/src/app/AccountApi.js
```diff
@@ -369,6 +369,7 @@ class AccountApi {
       if (e instanceof UnauthorizedError) {
         session.resetAndRedirect(routes.login);
       }
+      return undefined;
```
binwiederhier, line 372: Why is this required?

web/src/app/errors.js
```diff
@@ -77,3 +46,35 @@ export class IncorrectPasswordError extends Error {
...
+  throw new Error(`Unexpected response ${response.status}`);
+};
+
+export const fetchOrThrow = async (url, options) => {
+  const response = await fetch(url, options);
+  if (response.status !== 200) {
+    await throwAppError(response);
+  }
+  return response; // Promise!
+};
```
binwiederhier, line 80: Can you elaborate on the re-ordering?

web/src/app/Connection.js
nimbleghost: This rule is generally just to avoid any possible gotchas with JS' hoisting behaviour. So the moving was to satisfy having everything defined before its use.

- <https://github.com/airbnb/javascript#hoisting>
- <https://eslint.org/docs/latest/rules/no-use-before-define>
- <https://developer.mozilla.org/en-US/docs/Glossary/Hoisting>

web/src/app/Connection.js
nimbleghost: https://github.com/airbnb/javascript#variables--unary-increment-decrement

web/src/app/AccountApi.js
nimbleghost: The rule is consistent returns - `undefined` was already implied as the default return type, but since other branches do explicitly return values, it wants you to also return explicitly here too.

web/src/app/errors.js
nimbleghost: Same as the other cases above ^

web/src/app/Connection.js
binwiederhier: Alright, I guess :-D

web/src/app/AccountApi.js
nimbleghost: This is correct:

> The modified version removes the await keyword, making it a regular function that returns a Promise. The Promise will resolve to the value returned by upsertBillingSubscription.

This is not:

> So, the two versions are not completely equivalent in terms of the return value. The version with await returns the resolved value directly, while the version without await returns a Promise that resolves to the value.

The gist of is is that async/await is just syntactic sugar, it doesn't change how the function behaves when called.

The reason why `return await` is not required is that returning a promise and returning the awaited value of a promise results in exactly the same thing, unless you wrap it with try/catch and want to handle the error here.

<https://eslint.org/docs/latest/rules/no-return-await>

web/src/components/SubscriptionPopup.jsx
```diff
@@ -241,8 +241,6 @@ const DisplayNameDialog = (props) => {
           inputProps={{
             maxLength: 64,
             "aria-label": t("display_name_dialog_placeholder"),
-          }}
-          InputProps={{
```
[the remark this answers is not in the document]
nihalgonsalves: ah, you're right. will use this solution: https://github.com/mui/material-ui/issues/10064#issuecomment-373190215

web/src/components/PublishDialog.jsx
```diff
@@ -171,34 +171,33 @@ const PublishDialog = (props) => {
 
   const checkAttachmentLimits = async (file) => {
     try {
-      const account = await accountApi.get();
-      const fileSizeLimit = account.limits.attachment_file_size ?? 0;
-      const remainingBytes = account.stats.attachment_total_size_remaining;
+      const apiAccount = await accountApi.get();
```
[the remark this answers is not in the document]
nimbleghost: This was to satisfy `no-shadow` as there is a variable above with the same name. It's usually a hotly contested rule haha, can also remove it if you have a strong opinion about it.

web/src/components/PublishDialog.jsx
```diff
@@ -374,23 +373,23 @@ const PublishDialog = (props) => {
                   "aria-label": t("publish_dialog_priority_label"),
                 }}
               >
-                {[5, 4, 3, 2, 1].map((priority) => (
+                {[5, 4, 3, 2, 1].map((priorityMenuItem) => (
```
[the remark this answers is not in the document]
nimbleghost: Same as above, only for not shadowing

web/src/components/Notifications.jsx
```diff
@@ -468,53 +514,9 @@ const UserAction = (props) => {
...
-const ACTION_LABEL_SUFFIX = {
-  [ACTION_PROGRESS_ONGOING]: " …",
-  [ACTION_PROGRESS_SUCCESS]: " ✔",
-  [ACTION_PROGRESS_FAILED]: " ❌",
-};
-
 const NoNotifications = (props) => {
   const { t } = useTranslation();
-  const shortUrl = topicShortUrl(props.subscription.baseUrl, props.subscription.topic);
+  const topicShortUrlResolved = topicShortUrl(props.subscription.baseUrl, props.subscription.topic);
```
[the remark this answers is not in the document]
nimbleghost: ^

web/src/components/App.jsx
```diff
@@ -30,11 +30,14 @@ export const AccountContext = createContext(null);
 
 const App = () => {
   const [account, setAccount] = useState(null);
+
+  const contextValue = useMemo(() => ({ account, setAccount }), [account, setAccount]);
```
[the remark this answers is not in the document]
nimbleghost: If you create an object its identity changes for every render, forcing child components to re-render every time. `useMemo` makes sure the object identity doesn't change unless one of the dependents (in the array) change. It was a specific warning from eslint for `useContext` specifically, but this is generally good to do with React when passing down props.

web/src/components/PublishDialog.jsx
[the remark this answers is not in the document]
binwiederhier: no-shadow is fine. Makes sense.

web/src/app/AccountApi.js
binwiederhier: I think I mostly get it. The link was really helpful too, thanks.

web/src/components/PublishDialog.jsx
```diff
@@ -824,8 +824,11 @@ const ExpandingTextField = (props) => {
...
         sx={{ width: `${textWidth}px`, borderBottom: "none" }}
         InputProps={{
-          style: { fontSize: theme.typography[props.variant].fontSize, paddingBottom: 0, paddingTop: 0 },
-          "aria-label": props.placeholder,
+          style: { fontSize: theme.typography[props.variant].fontSize },
+          inputProps: {
+            style: { paddingBottom: 0, paddingTop: 0 },
+            "aria-label": props.placeholder,
+          },
         }}
```
binwiederhier, line 830: They should be adjacent, not nested:

```
        InputProps={{
          style: { fontSize: theme.typography[props.variant].fontSize },
        }}
        inputProps={{
          style: { paddingBottom: 0, paddingTop: 0 },
          "aria-label": props.placeholder,
        }}
```

web/src/components/Account.jsx
```diff
@@ -1008,8 +1003,6 @@ const TokenDeleteDialog = (props) => {
       console.log(`[Account] Error deleting token`, e);
       if (e instanceof UnauthorizedError) {
         session.resetAndRedirect(routes.login);
-      } else {
-        setError(e.message);
       }
```
binwiederhier, line 1006: This should be `<DialogFooter status={error}>`, instead of deleting it

nimbleghost:
Thanks for the review and final fixes! The `inputProps` being nested is apparently an equivalent workaround for the lint rule conflicting: https://github.com/mui/material-ui/issues/10064#issuecomment-373190215 but disabling it is also fine 👍

PR #749, opened 2023-05-24
Switch to Vite

nimbleghost:
This is baed on #748, so should either probably be rebased and reviewed after merging that. It's not that simple to do chained PRs across forks haha.

Actual diff: https://github.com/nimbleghost/ntfy/compare/add-eslint...switch-to-vite?expand=1


binwiederhier:
I "rebased" and tried this out locally and here https://github.com/binwiederhier/ntfy/pull/752, and I got the same error when building. 

I had to re-add emotion for it to work

```
    "@emotion/react": "^11.11.0",
    "@emotion/styled": "^11.11.0",
```

PR #751, opened 2023-05-24
Add PWA manifest, service worker, and web push

### LINKED ISSUE #199: Push API support
This is a very neat tool. It's missing one feature that would make it a fantastic replacement for AWS/Azure alternative, and that is Push API support https://developer.mozilla.org/en-US/docs/Web/API/Push_API.

Any planes on enabling the server to push events to web clients? The goal is to get around the current limitation of the UI client, that needs to remain open to receive events.

### LINKED ISSUE #346: Support Web Push + PWA, for background notifications
I am not sure what running as progressive web app requires, but I understand they should be able to support receiving notifications in the background even without the tab/app open as long as browser is running.

I think currently at least webapp manifest is missing.

* https://developer.mozilla.org/docs/Web/Manifest
* https://developer.mozilla.org/docs/Web/Progressive_web_apps
* https://caniuse.com/web-app-manifest (add to homescreen)

~~Somewhat unrelatedly, iOS is also receiving support for web push sometime next year, https://webventures.rejh.nl/blog/2022/06/25/the-biggest-thing-from-wwdc22/~~

Update 2023-04-02: iOS 16.4 supports web push when the PWA has been added to the homescreen.

nimbleghost:
Resolves #346
Resolves #199

Adds a PWA manifest, service worker, the required go server routes, and VAPID-based web push

I can't test Android, but this is what it looks like on macOS and iOS:

| | |
|--|--|
| chrome_install_prompt | chrome_installed |

| | | |
|--|--|--|
| ios_add | ios_opts | iOS-subscribe |
| iOS-notif-1 | ios_badge | ios_opts |


server/server.go
```diff
@@ -76,6 +76,9 @@ var (
 	publishPathRegex       = regexp.MustCompile(`^/[-_A-Za-z0-9]{1,64}/(publish|send|trigger)$`)
 
 	webConfigPath                                        = "/config.js"
+	webManifestPath                                      = "/manifest.webmanifest"
+	webServiceWorkerPath                                 = "/sw.js"
+	webRegisterServiceWorkerPath                         = "/registerSW.js"
```
binwiederhier, line 81: Could these just go in `/static/` somewhere? That'd make it easier, since we wouldn't need special routes for them. We could move in the build process.

nimbleghost: I believe the manifest has to be at that path. The service workers could possibly be moved, though I don't really want to configure it differently since there are a bunch of `scope` related gotchas that I don't fully understand yet - I think it could possibly limit the service worker to a sub path.

nimbleghost:
> I always assumed that ntfy will be a push provider, but my understanding is that that's not possible.

It does work! I have it working with VAPID keys and no intermediate (e.g. Firebase) required. Pushing now - I will send you a message on Discord to discuss how to configure it though, that's what I'm looking at currently.

server/server.go
nimbleghost: I checked - I removed registerSW by inlining it, but both the manifest and sw.js have to be at the root.

server/server.go
```diff
@@ -877,6 +977,96 @@ func (s *Server) forwardPollRequest(v *visitor, m *message) {
...
+
+	if err != nil {
+		logvm(v, m).Err(err).Warn("Unable to publish web push messages")
+		return
+	}
+
+	totalCount := len(subscriptions)
+
+	wg := &sync.WaitGroup{}
+	wg.Add(totalCount)
```
binwiederhier, line 991: Remove the wait group. The wait group is not used. You never call `wg.Wait` anywhere, though sinc it's fire and forget, you don't need to anyway

server/server.go
```diff
@@ -877,6 +977,96 @@ func (s *Server) forwardPollRequest(v *visitor, m *message) {
 	}
 }
 
+func (s *Server) publishToWebPushEndpoints(v *visitor, m *message) {
+	subscriptions, err := s.webPushSubscriptionStore.GetSubscriptionsForTopic(m.Topic)
+
```
binwiederhier, line 982: 1. remove new line
2. Idiomatic Go doesn't have "Get" Prefix, so it'd be `webPush.SubscriptionsForTopic(m.Topic)` here

server/server.go
```diff
@@ -877,6 +977,96 @@ func (s *Server) forwardPollRequest(v *visitor, m *message) {
...
+				err = s.webPushSubscriptionStore.ExpireWebPushEndpoint(sub.BrowserSubscription.Endpoint)
+				if err != nil {
+					logvm(v, m).Err(err).Fields(ctx).Warn("Unable to expire subscription")
+				}
+
+				return
+			}
+		}(i, xi)
+	}
+}
```
binwiederhier, line 1068: I have many opinions (mostly stylistic) around this function. I think I should just modify it myself. Can I somehow contribute to your branch?

In short:
- This should also move to `server_webpush.go`
- I don't like the `log..(); return` pattern. If the function is long enough; I usually create a `func publishToWebPushEndpointsInternal(...) error` function that can `return err`, and then have the actual `publishToWebPushEndpoints` function do the `log...() return` thingy
- I'm not sure if I like the emoji formatting thing on the server. How much hassle is it to do it in the sw?

server/server.go
```diff
@@ -877,6 +977,96 @@ func (s *Server) forwardPollRequest(v *visitor, m *message) {
...
+				return
+			}
+
+			resp, err := webpush.SendNotification(jsonPayload, &sub.BrowserSubscription, &webpush.Options{
+				Subscriber:      s.config.WebPushEmailAddress,
+				VAPIDPublicKey:  s.config.WebPushPublicKey,
+				VAPIDPrivateKey: s.config.WebPushPrivateKey,
+				// deliverability on iOS isn't great with lower urgency values,
+				// and thus we can't really map lower ntfy priorities to lower urgency values
+				Urgency: webpush.UrgencyHigh,
```
binwiederhier, line 1041: I vaguely remember reading about Apple penalizing you if you only send high urgency alerts. Not sure if true though, but it may be worth checking again

server/server.yml
```diff
@@ -38,6 +38,16 @@
...
 
+# Enable web push
+#
+# Run ntfy web-push generate-keys to generate the keys
+#
+# web-push-enabled: true
+# web-push-public-key: ""
+# web-push-private-key: ""
+# web-push-subscriptions-file: ""
+# web-push-email-address: ""
```
binwiederhier, line 49: - This needs to be a little more verbose. Usually this section is largely identical to whatever I put in the config.md section on the topic.
- The commented values should match the defaults, so in this case:

```yaml
# web-push-enabled: false
# web-push-public-key: 
# web-push-private-key: 
# web-push-subscriptions-file: 
# web-push-email-address: 
```

server/web_push.go
```diff
@@ -0,0 +1,132 @@
...
+	deleteWebPushSubscriptionByEndpointQuery         = `DELETE FROM web_push_subscriptions WHERE endpoint = ?`
+	deleteWebPushSubscriptionByUsernameQuery         = `DELETE FROM web_push_subscriptions WHERE username = ?`
+	deleteWebPushSubscriptionByTopicAndEndpointQuery = `DELETE FROM web_push_subscriptions WHERE topic = ? AND endpoint = ?`
+
+	selectWebPushSubscriptionsForTopicQuery = `SELECT endpoint, key_auth, key_p256dh, username FROM web_push_subscriptions WHERE topic = ?`
+
+	selectWebPushSubscriptionsCountQuery = `SELECT COUNT(*) FROM web_push_subscriptions`
+)
+
+type webPushSubscriptionStore struct {
```
binwiederhier, line 40: I'd call this webPushStore, probably. But that's a personal preference.

server/web_push.go
```diff
@@ -0,0 +1,132 @@
...
+	"database/sql"
+
+	_ "github.com/mattn/go-sqlite3" // SQLite driver
+)
+
+// Messages cache
+const (
+	createWebPushSubscriptionsTableQuery = `
+		BEGIN;
+		CREATE TABLE IF NOT EXISTS web_push_subscriptions (
```
binwiederhier, line 13: 1. Maybe just call the table `subscriptions`, since the entire database is about web push?
2. The updated_at column is never used
3. When do subscriptions expire, or how are they kept alive? To prevent infinite growth of this table, we have to actively prune records. I'd be okay with having a long period before pruning, but at some point, an unused record has to be deleted. Imaging thousands of people trying it out and then never using it again.

web/public/config.js
```diff
@@ -7,7 +7,7 @@
 
 var config = {
   base_url: window.location.origin, // Change to test against a different server
-  app_root: "/app",
+  app_root: "/",
```
binwiederhier, line 10: Good change. I wonder how long it took you to figure this one out. I even stumble over this occasionally. "WHY IS THE '/' PAGE BLANK???"

nimbleghost:
thanks for the review and good questions. I'll respond in detail when I have some free time or after work.

> (I'm going to have a lot of nit-picky responses. Please do know that I am extremely impressed by what you have delivered here. But I am a little OCD about this stuff.)

much appreciated, I'm only vaguely familiar with go so I'm learning the patterns / idiomatic way of doing things as I go (haha).

docs/subscribe/desktop.md
```diff
@@ -0,0 +1,12 @@
+# Using the web app as an installed PWA
```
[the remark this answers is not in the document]
nimbleghost: I suppose - just thought it would be nice to highlight that you can have an individual app in your dock/etc by using Chrome or Edge, including receiving background notifications as long as either the PWA or browser is running. There isn't much functional difference from an Electron app for an app like ntfy.

server/server.go
```diff
@@ -32,50 +25,65 @@ import (
...
+	smtpServerBackend        *smtpBackend
+	smtpSender               mailer
+	topics                   map[string]*topic
+	visitors                 map[string]*visitor // ip:<ip> or user:<user>
+	firebaseClient           *firebaseClient
+	messages                 int64                               // Total number of messages (persisted if messageCache enabled)
+	messagesHistory          []int64                             // Last n values of the messages counter, used to determine rate
+	userManager              *user.Manager                       // Might be nil!
+	messageCache             *messageCache                       // Database that stores the messages
+	webPushSubscriptionStore *webPushSubscriptionStore           // Database that stores web push subscriptions
```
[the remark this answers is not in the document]
nimbleghost: (the library is also just called `webpush`, but I guess it isn't that confusing if you're accessing via `s.webPush`)

server/server.go
```diff
@@ -535,6 +570,63 @@ func (s *Server) handleTopicAuth(w http.ResponseWriter, _ *http.Request, _ *visi
...
+		return err
+	}
+
+	err = s.webPushSubscriptionStore.AddSubscription(topic.ID, username, sub)
+	if err != nil {
+		return err
+	}
+
+	return s.writeJSON(w, newSuccessResponse())
+}
```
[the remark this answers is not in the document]
nimbleghost: I did already wrap this with `limitRequestsWithTopic`, but I haven't looked at the details of how that works. Is that not enough here?

I also wonder if we should limit the endpoints to known browser endpoints as an additional measure.

server/server.go
nimbleghost: I can take a stab at improving this but you can definitely push to my branch, though I do often rebase / squash / fixup / amend while working, so if you want to I can push all my current stuff and not do that until you're done:

CleanShot 2023-05-30 at 12 37 29@2x

r.e. emoji formatting: initially I thought I couldn't do imports easily without a lot of transpile config tinkering, but it does now work seamlessly. only concern is service worker size, but the emoji file is small enough and loaded in the browser anyway. I'll move it.

server/server.go
nimbleghost: I had this running with a different mapping before, and lower urgency alerts would just not come through. perhaps it also interacts with TTL - maybe setting a higher TTL would result in lower urgency alerts coming in after a device is connected to power/wifi and able to receive lower urgencies again - but needs testing.

fwiw in the last couple of days I haven't had Apple block any of the high urgency alerts. I would like to do some further testing with normal urgency/longer TTLs/etc but I wouldn't change this for now.

Urgency reference: https://datatracker.ietf.org/doc/html/rfc8030#section-5.3

nimbleghost: Safari's docs: https://developer.apple.com/documentation/usernotifications/sending_web_push_notifications_in_web_apps_safari_and_other_browsers

> To indicate to the push service whether to send the notification immediately or prioritize the recipient’s device power considerations for delivery, provide one of the following values: very-low, low, normal, or high. To attempt to deliver the notification immediately, specify high.

server/web_push.go
nimbleghost: I put the `updated_at` column in to have the possibility to add expiry logic in the future, and to do some manual testing. I'm not so sure how it should work yet.

If a user revokes permission, the push service sends a `410 Gone` which will cause our publisher to delete the subscription automatically. In practice this isn't always the case - it works immediately for Chrome, but at least Apple doesn't seem to do it.

I also added a scheduled update in the web app, so people who regularly open the web app should have a recent `updated_at` - the real problem is when people subscribe to something and then don't open ntfy. It would not be the best UX if subscriptions silently expire - so I wonder if the solution is to send a message such as "Please reopen the ntfy web app to continue receiving background notifications" after e.g. 1 month, then expire a day later.

Sort of tangentially, we could also have the service worker notify the ntfy server about receipt of a web push event and store that (e.g. some sort of `last_sent_at` and `last_received_by_web_app_at` representation). There's also a spec for getting receipts, but I don't see any implementations for it, and the service worker reporting it would be a more complete guarantee anyway https://datatracker.ietf.org/doc/html/rfc8030#section-6.3

docs/subscribe/desktop.md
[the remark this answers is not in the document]
binwiederhier: I'd argue that the difference is the executable / installer. I very much do think there is a difference between the "Slack" I install and the Slack web app. Even though it's 99% the same. I don't care enough though. We can leave it like this, except for the "PWA" phrase in the title. I don't even know what that means, and neither does a user.

server/server.go
binwiederhier: > fwiw in the last couple of days I haven't had Apple block any of the high urgency alerts. 

We shall see :-D It's never a problem at small scale, but may become one if they penalize you if all of them are urgent. I bet it'll be fine though.

nimbleghost: Yeah, let's leave it and do separate testing on TTLs and urgency. It's a shame Apple doesn't go into more detail; testing battery and network dependent things is the worst.

server/web_push.go
binwiederhier: > I also added a scheduled update in the web app, so people who regularly open the web app should have a recent updated_at 

:+1: 

> the real problem is when people subscribe to something and then don't open ntfy.

Don't you think people would click on the notification every now and then, and that would bring them to the web app, which would then send a "I'm still alive" message? This is how I keep access tokens for the web session alive. They expire after 72 hours, but are always extended by 72 hours unless you log in. 

Tangentially (a lot of tangents...), does this access token logic^ cause issues with the PWA?

> It would not be the best UX if subscriptions silently expire - so I wonder if the solution is to send a message such as "Please reopen the ntfy web app to continue receiving background notifications" after e.g. 1 month, then expire a day later.

Assuming there are truly scenarios in which people never ever open the web app, then the UX would be bad. But don't you think people will click on the notification?

docs/subscribe/desktop.md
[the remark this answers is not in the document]
nimbleghost: Agreed on the wording! There is an "executable", i.e. it creates an "Application" on macOS that is completely separate from Chrome, not sure what it looks like on Linux & Windows.

In the end it all comes down to what ntfy users would want from a desktop app that a PWA couldn't provide. for e.g. I was wondering what we could do with installable PWAs – unfortunately `ntfy://` won't be possible - but registering `web+ntfy://` would 🤔 [Web/API/Navigator/registerProtocolHandler](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/registerProtocolHandler).

server/web_push.go
binwiederhier: I hate any sort of user tracking. ;-) This doesn't really serve a purpose. 

Eventually when we do a synchronized "mark as read" then it'll serve a purpose, but now if it's just for tracking, then i don't like that

nimbleghost: > I hate any sort of user tracking. ;-) This doesn't really serve a purpose.

Fair, though I mean tracking the service worker receiving the message - so ACK not tracking whether the user interacts or reads with it at all.

server/server.go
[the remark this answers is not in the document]
binwiederhier: > I did already wrap this with limitRequestsWithTopic, but I haven't looked at the details of how that works. Is that not enough here?

`limitRequestsWithTopic` limits the requests in total, yes. On ntfy.sh this is 250 a day, but on a default instance it's ~17k per day. So I think something needs to be limited here.  *The rule of the Internet is: What can be abused, will be abused.*

I don't have a great answer as to how the limit should look like. The `visitor` already has like ~10 limiters for various actions, but I am always hesitant to add more, because every limiter we adds more RAM usage...

The only sensible option I can think of is this: Add `webPushSubscriptionLimiter *util.RateLimiter` with a fixed rate (not configurable, e.g. 10 per 3 hours or something like that), and only initialize the limiter if s.WebPushEnabled is true

> I also wonder if we should limit the endpoints to known browser endpoints as an additional measure (edit: https://github.com/pushpad/known-push-services/blob/master/whitelist)

This is to limit outgoing requests, right? That seems like the wrong thing to limit. If anything, I'd add 429-handling somehow, but I prefer limiting as early as possible (on the incoming end).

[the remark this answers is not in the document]
nimbleghost: > The only sensible option I can think of is this: Add webPushSubscriptionLimiter *util.RateLimiter with a fixed rate (not configurable, e.g. 10 per 3 hours or something like that), and only initialize the limiter if s.WebPushEnabled is true

I guess it's also a balance between how often to update the frontend. I set it to every 5 minutes, but can be much slower - the question is whether it would ever run on the mobile use case of looking a notification and closing the app again (except for the startup refresh). So I wonder if I should have it set to check every 5 minutes plus every time the window comes into the foreground (visibilityState), but no-op if the web push was updated was in the last ~x hours.

10 per 3 hours still seems kinda low nevertheless - or is it per topic?

> This is to limit outgoing requests, right? That seems like the wrong thing to limit. If anything, I'd add 429-handling somehow, but I prefer limiting as early as possible (on the incoming end).

Yes, I was thinking of it as an additional check to prevent using a ntfy server to DDoS another endpoint. There doesn't seem to be a good reason to allow non-known browser push services.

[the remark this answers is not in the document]
nimbleghost: > I'd add 429-handling somehow

This is already done, any non 2xx status from the web push endpoint will delete the subscription.

binwiederhier:
Did some work. Not even close to done.

- https://github.com/binwiederhier/ntfy/pull/751/commits/8f2120e86707f18190815b1d2c14b40c577f8df8 changes some names
- https://github.com/binwiederhier/ntfy/pull/751/commits/0d2f0dd3f693b2f43551a543126d1528a3ee5c8c moves stuff to a new file
- https://github.com/binwiederhier/ntfy/pull/751/commits/ad761f4492e82712f5898ba045a95ed1cabb1654 makes things more Go-y, and switches to `user_id` in the database
- https://github.com/binwiederhier/ntfy/pull/751/commits/35ac0956a81a646214449b1030dd0d4a79e2c132 moves the config to config.js

Things I've noticed:
- I love that the subscription drop down now has icons. Random change, but love it!
- I do not for the life of me understand what "Sound only" means. I didn't look at the code, but from a user perspective, that's odd
- I have that the "SubscribeDialog" now has 4(!) toggles. I don't think I want to even give people the option to opt-out of desktop notifications. If they want to, they can just mute the topic. That way, it'll be just 3 options (or we could even drop the web-push one and just make it available in the drop down, idk yet.

Lots of TODOs. Here are some major things:
- [ ] The subscribe/unsubscribe endpoints should be POST + DELETE to the same endpoint v1/webpush/$topic or something
- [ ] The unsubscribe endpoint does not work at all for me, the payload is empty

(I'll do more tonight, likely)

nimbleghost:
> I love that the subscription drop down now has icons. Random change, but love it!

:D added the checkmarks for the type and then the rest of the list looked empty

---

> I do not for the life of me understand what "Sound only" means. I didn't look at the code, but from a user perspective, that's odd

fair. the reasoning was that if a user denied browser notification access at all, I wanted there to be a way to represent that. or if the browser/context doesn't support it, e.g. http or opening a site in incognito. how would you represent that? It's distinct from muting the topic; I wouldn't want to force a topic muted state if the notification APIs were unavailable/denied. Maybe "No notifications"?

---

> I have that the "SubscribeDialog" now has 4(!) toggles. I don't think I want to even give people the option to opt-out of desktop notifications. If they want to, they can just mute the topic. That way, it'll be just 3 options (or we could even drop the web-push one and just make it available in the drop down, idk yet.

Hmm, the desktop notification toggle was again browser notification opt-in UX driven. When you first visit the website and click subscribe, the toggle defaults to off:

CleanShot 2023-05-30 at 21 17 19@2x

You tap it, then you either click allow and it goes to enabled, or you see this flow if denied or unavailable (browsers will silently deny in incognito for example):

CleanShot 2023-05-30 at 21 18 17@2x

Maybe the solution is to have a little dialog that shows up when you subscribe for the first time, asking to enable notifications first. I do agree that there are too many toggles (esp. when logged in).

---

relatedly, I'm wondering if the "reserve topic" toggle could be replaced with just the dropdown with a default "public" setting.

---

> The subscribe/unsubscribe endpoints should be POST + DELETE to the same endpoint v1/webpush/$topic or something

that's what I wanted initially, but you can't have a DELETE body, and we need to specify the endpoint we want to unsubscribe. since it's a URL it's inconvenient to embed in a URL, or it could be `DELETE /topic/webpush/<subscription-id>`.. 

>  The unsubscribe endpoint does not work at all for me, the payload is empty

which payload? it works for me:

```sh
$ curl https://<server>/test/web-push/unsubscribe \
  -d '{"endpoint":"https://fcm.googleapis.com/fcm/send/<xyz>"}' \
  -H "Authorization: Bearer xxx"

{"success":true}
```

nimbleghost:
Alright, addressed most of the review comments. What's remaining:

- Naming: "Sound only"
- UX around browser permission approval and the 4 subscribe toggles
- Rate limit subscribe?
- Perhaps limit publish to known browser push endpoints (https://github.com/pushpad/known-push-services/blob/master/whitelist)

Do you see anything else?

binwiederhier:
Additional action items:

- "Bulk endpoints"?
- Changing endpoints to v1/account/webpush
- Subscription expiration (+ warning notification)

I'll try to actually get some time to try it tomorrow and on the weekend.

server/config.go
```diff
@@ -22,6 +23,12 @@ const (
 	DefaultStripePriceCacheDuration             = 3 * time.Hour    // Time to keep Stripe prices cached in memory before a refresh is needed
 )
 
+// Defines default web push settings
+const (
+	DefaultWebPushExpiryWarningDuration = 7 * 24 * time.Hour
+	DefaultWebPushExpiryDuration        = DefaultWebPushExpiryWarningDuration + 24*time.Hour
+)
```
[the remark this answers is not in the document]
nimbleghost: *shorter

web/src/components/SubscriptionPopup.jsx
```diff
@@ -267,6 +334,48 @@ const DisplayNameDialog = (props) => {
...
+
+  return (
+    <MenuItem>
+      {subscription.webPushEnabled === 1 && checkedItem}
+      <ListItemText inset={subscription.webPushEnabled !== 1} onClick={handleToggleBackground}>
+        {t("notification_toggle_background")}
+      </ListItemText>
+    </MenuItem>
+  );
+};
```
binwiederhier, line 377: As discussed in Discord, I looove the idea of just having one global config option `Background notifications: Yes|No` in the Settings and remove all of these per-subscription toggles. 

That would remove
- The dropdown menu item "Background notifications" for each subscription
- The additional switch in the Subscribe dialog

It would be a local-only setting (for now).

Behavior:
- When enabled (`Background notifications: Yes`), the `v1/webpush` endpoint would be called with all subscribed topics listed.
- When disabled (`Background notifications: No`),  the `v1/webpush` endpoint would be called with an empty array.
- When a new topic is subscribed to (in the SubscribeDialog) and background notifications are enabled, the `v1/webpush` endpoint is called
- On login, background notifications are (for now) not enabled. This can change later when server-side per-device[-type] settings are possible
- On logout, the `v1/webpush` endpoint would be called with an empty array.

Good?

web/src/components/Preferences.jsx
```diff
@@ -242,19 +242,14 @@ const WebPushEnabled = () => {
...
-
   if (!notifier.pushPossible()) {
     return null;
   }
 
   return (
     <Pref labelId={labelId} title={t("prefs_notifications_web_push_title")} description={t("prefs_notifications_web_push_description")}>
       <FormControl fullWidth variant="standard" sx={{ m: 1 }}>
-        <Select value={defaultEnabled} onChange={handleChange} aria-labelledby={labelId}>
+        <Select value={defaultEnabled ?? false} onChange={handleChange} aria-labelledby={labelId}>
```
nimbleghost: Won't this make the UI show "Disabled" when loading, even if the preference is true?

web/src/components/SubscriptionPopup.jsx
```diff
@@ -207,29 +208,41 @@ export const SubscriptionPopup = (props) => {
...
+          </MenuItem>
+        )}
+        {!subscription.mutedUntil && (
+          <MenuItem onClick={() => handleSetMutedUntil(1)}>
+            <ListItemIcon>
+              <NotificationsOff fontSize="small" />
+            </ListItemIcon>
+            {t("action_bar_mute_notifications")}
+          </MenuItem>
+        )}
```
nimbleghost, line 241: I would do `{subscription.mutedUntil ? ( ... ) : ( ... )}`. The `!!` is unnecessary since the ternary will coerce to Boolean.

server/server_web_push.go
```diff
@@ -30,109 +30,98 @@ var webPushEndpointAllowRegex = regexp.MustCompile(webPushEndpointAllowRegexStr)
...
+	for _, subscription := range subscriptions {
+		ctx := log.Context{"endpoint": subscription.BrowserSubscription.Endpoint, "username": subscription.UserID, "topic": m.Topic, "message_id": m.ID}
+		if err := s.sendWebPushNotification(payload, subscription, &ctx); err != nil {
+			log.Tag(tagWebPush).Err(err).Fields(ctx).Warn("Unable to publish web push message")
+		}
 	}
 }
 
+// TODO this should return error
+// TODO the updated_at field is not actually updated ever
```
nimbleghost, line 79: It does get updated: https://github.com/binwiederhier/ntfy/pull/751/files#diff-dce7312bf864bc651415e35f78f59fd4d8d5dbdeea77c005f9042b206a8014edR30-R32

because we use `INSERT OR REPLACE INTO` on both create/update, the `updated_at` is in fact updated, using the default value of the column, the current timestamp: `updated_at DATETIME DEFAULT CURRENT_TIMESTAMP`

server/server_web_push.go
```diff
@@ -30,109 +30,98 @@ var webPushEndpointAllowRegex = regexp.MustCompile(webPushEndpointAllowRegexStr)
...
+	payload, err := json.Marshal(newWebPushPayload(fmt.Sprintf("%s/%s", s.config.BaseURL, m.Topic), m))
+	if err != nil {
+		log.Tag(tagWebPush).Err(err).Warn("Unable to marshal expiring payload")
+		return
+	}
+	for _, subscription := range subscriptions {
+		ctx := log.Context{"endpoint": subscription.BrowserSubscription.Endpoint, "username": subscription.UserID, "topic": m.Topic, "message_id": m.ID}
+		if err := s.sendWebPushNotification(payload, subscription, &ctx); err != nil {
+			log.Tag(tagWebPush).Err(err).Fields(ctx).Warn("Unable to publish web push message")
+		}
```
nimbleghost: Doesn't this run sequentially now, without the `go func`?

web/src/components/Preferences.jsx
binwiederhier: Yes, but only very shortly. I have to figure the whole loading thing out overall anyway. It's very ugly to have a two DOM changes in succession that change the height of components. It's very choppy looking.

It's just personal preference to do it this way. Plus, false is the default.

web/src/components/SubscriptionPopup.jsx
binwiederhier: A ternary for long stuff like this is very hard to read. And readability is king king king 👑

server/server_web_push.go
binwiederhier: Oh cool. And that does get triggered regularly from the worker?

server/server_web_push.go
binwiederhier: It does, but the whole thing runs in a go routine. No need to spin up more and make the code less readable IMHO.

server/server_web_push.go
nimbleghost: yup exactly 👌

web/src/components/SubscriptionPopup.jsx
nimbleghost: style preference I guess, but I find the way prettier formats the jsx ternaries pretty readable. 

anyway `&&` also coerces to boolean, so personally would just do `{subscription.mutedUntil && (...)}` then because double negation always makes me do a double take, but this is just nitpicking

server/server_web_push.go
nimbleghost: fair enough, I'm just imagining if there were hundreds of web push subscriptions on a topic, this would take some time to send them all, vs running all web push sends in parallel. and it would also make the publisher wait, I believe?

binwiederhier:
Todo list:
- [ ] Add rate limiting somehow
- [ ] Change the webpush endpoint JSON format
- [ ] Change the updated_at to an INT (like all timestamps in other databases)
- [ ] Split ExpireAndGetExpiringSubscriptions into two functions
- [ ] Add a schemaVersion table
- [ ] Collapse UpdateSubscriptions to actually make it a transaction
- [ ] Convert all non-pointer params to pointer params, e.g. []webPushSub -> []*webPushSub

nimbleghost:
>  Revert blue "Click here alert" to yellow one

Not a problem, but the reasoning was that it shows up on first visit, so more a gentle request. The "permission denied" or "not supported" are still a warning yellow.

web/src/components/Navigation.jsx
```diff
@@ -344,16 +353,46 @@ const SubscriptionItem = (props) => {
...
-      <Alert severity="warning" sx={{ paddingTop: 2 }}>
-        <AlertTitle>{t("alert_grant_title")}</AlertTitle>
-        <Typography gutterBottom>{t("alert_grant_description")}</Typography>
-        <Button sx={{ float: "right" }} color="inherit" size="small" onClick={props.onRequestPermissionClick}>
-          {t("alert_grant_button")}
-        </Button>
+      <Alert severity="info" sx={{ paddingTop: 2 }}>
+        <AlertTitle>{t("alert_notification_ios_install_required_title")}</AlertTitle>
+        <Typography gutterBottom>{t("alert_notification_ios_install_required_description")}</Typography>
       </Alert>
```
binwiederhier, line 396: If I understand correctly, this will show up for all iOS devices and it will persist unless you do what the description says. What if somebody just wants to use the ntfy web app with web sockets and tab open? Will this alert stay there? 

If this is the case, we should add a [Dismiss] button, right?

nimbleghost: That's correct, but it's similar to the current implementation, it shows a persistent "not supported" on iOS:

15B87DDD-7702-466F-A18B-85E014055968_1_105_c

So this change would basically change it to something the user can fix if wanted.

web/public/sw.js
```diff
@@ -0,0 +1,180 @@
...
+precacheAndRoute(self.__WB_MANIFEST);
+
+// clean old assets
+cleanupOutdatedCaches();
+
+// to allow work offline
+if (import.meta.env.MODE !== "development") {
+  registerRoute(new NavigationRoute(createHandlerBoundToURL("/")));
+  registerRoute(({ url }) => url.pathname === "/config.js", new NetworkFirst());
+}
```
binwiederhier, line 180: This entire file is a bit messy: Lots of nested, unnamed, async functions, and very little explanation. 

If the service worker supports it, can we 
- break out some functions (could also be inside the `addEventListener`)
- add some docs to the file itself and to the individual functions
?

I can probably do that too, but I think I have to stop working on this for now. I think given that service workers are less common knowledge, it's important to document well what's happening here.

web/public/sw.js
```diff
@@ -0,0 +1,180 @@
...
+            timestamp: message.time * 1_000,
+            actions,
+          }),
+        ]);
+      } else {
+        // We can't ignore the push, since permission can be revoked by the browser
+        await self.registration.showNotification("Unknown notification received from server", {
+          body: "You may need to update ntfy by opening the web app",
+          icon: "/static/images/ntfy.png",
+          data,
```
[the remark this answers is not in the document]
nimbleghost: that issue is unrelated, the `i18n` library we use is already service-worker compatible. fixed!

web/public/sw.js
nimbleghost: Done, let me know what's still unclear or confusing

web/src/app/Notifier.js
```diff
@@ -1,39 +1,34 @@
...
+
+    const registration = await this.serviceWorkerRegistration();
+    await registration.showNotification(
+      ...getNotificationParams({
+        subscriptionId: subscription.id,
+        message: notification,
+        defaultTitle,
+        topicRoute: new URL(routes.forSubscription(subscription), window.location.origin).toString(),
+      })
+    );
```
binwiederhier, line 31: If I understand correctly, this now implies that the service worker is correctly registered to display notifications in the normal web app. What is this change about? Just to harmonize things? What is the risk here?

I sadly cannot test half the features, because I'm on Linux and there are no action buttons and such.

nihalgonsalves: yes - at first I left it alone but I've been testing it over the last few weeks and the sw setup is stable.

- The main advantage is making them work the same and not diverge in behaviour. 
- The downside is not working where service workers do not work, but that's not really a risk: https://caniuse.com/serviceworkers - the one thing that I've seen is that Firefox Private Mode doesn't support them, but ntfy currently doesn't work there anyway due to IndexedDB.

P.S. you can test actions on Chrome, on Linux too: https://developer.mozilla.org/en-US/docs/Web/API/Notification/actions#browser_compatibility

nimbleghost: yes - at first I left it alone but I've been testing it over the last few weeks and the sw setup is stable.

- The main advantage is making them work the same and not diverge in behaviour.
- The downside is not working where service workers do not work, but that's not really a risk: <https://caniuse.com/serviceworkers> - the one thing that I've seen is that Firefox Private Mode doesn't support them, but ntfy currently doesn't work there anyway due to IndexedDB.

P.S. you can test actions on Chrome, on Linux too: <https://developer.mozilla.org/en-US/docs/Web/API/Notification/actions#browser_compatibility>

web/src/app/notificationUtils.js
```diff
@@ -0,0 +1,70 @@
...
+  return m.message;
+};
+
+const isImage = (filenameOrUrl) => filenameOrUrl?.match(/\.(png|jpe?g|gif|webp)$/i) ?? false;
+
+export const icon = "/static/images/ntfy.png";
+export const badge = "/static/images/mask-icon.svg";
+
+export const toNotificationParams = ({ subscriptionId, message, defaultTitle, topicRoute }) => {
+  const image = isImage(message.attachment?.name) ? message.attachment.url : undefined;
```
nimbleghost, line 43: do we maybe want to do `attachment?.type?.startsWith('image/') || attachment?.filename?.match(...)`?

server/webpush_store.go
```diff
@@ -2,15 +2,23 @@ package server
...
 
 	_ "github.com/mattn/go-sqlite3" // SQLite driver
 )
 
 const (
-	subscriptionIDPrefix = "wps_"
-	subscriptionIDLength = 10
+	subscriptionIDPrefix             = "wps_"
+	subscriptionIDLength             = 10
+	subscriptionLimitPerSubscriberIP = 10
```
nimbleghost, line 16: [nit] `subscriptionEndpointLimit...`? It wasn't clear at first that it was per endpoint and not per topic.

server/webpush_store.go
```diff
@@ -2,15 +2,23 @@ package server
...
+)
+
+var (
+	errWebPushNoRows               = errors.New("no rows found")
```
nimbleghost, line 20: I'm not sure I understand this error – won't there always be a row if you do `count`?

server/server_web_push.go
binwiederhier: The `publishToWebPushEndpoints` function is called in two places, and it's only called as `go publishToWebPushEndpoints()`, so the publisher doesn't have to wait. And usually there are only 1-2 subscribers. There are only a handful of topics with more subscribers.

server/webpush_store.go
binwiederhier: It's just defensive programming. Handle every possible error. In this case, we handle "!rows.Next()" even though it should never happen.

server/webpush_store.go
binwiederhier: Changed.

web/src/app/Notifier.js
binwiederhier: Thanks for the explanation! I think it's fine. Let's see if we break anything :grimacing:

server/server_account_test.go
```diff
@@ -431,6 +431,41 @@ func TestAccount_Delete_Not_Allowed(t *testing.T) {
 	require.Equal(t, 40026, toHTTPError(t, rr.Body.String()).Code)
 }
 
+func TestAccount_Delete_Success_WithWebPush(t *testing.T) {
```
nimbleghost, line 434: There's already `TestServer_WebPush_DeleteAccountUnsubscribe` in the `sever_webpush_test.go`, may have put it in the wrong place haha

server/server_webpush_test.go
```diff
@@ -0,0 +1,256 @@
...
+	require.Equal(t, `{"success":true}`+"\n", response.Body.String())
+
+	requireSubscriptionCount(t, s, "test-topic", 1)
+
+	request(t, s, "DELETE", "/v1/account", `{"password":"ben"}`, map[string]string{
+		"Authorization": util.BasicAuth("ben", "ben"),
+	})
+	// should've been deleted with the account
+	requireSubscriptionCount(t, s, "test-topic", 0)
+}
```
nimbleghost, line 146: (here)

server/server_account_test.go
binwiederhier: I'll remove the other one

PR #754, opened 2023-05-26
Add a way to use Docker for building everything

nimbleghost:
Add a way to use Docker for building everything

I’d like to test #751 on my own instance, but installing all the build
dependencies on my server isn’t ideal - having this script in the repo
would make it possible to simply point my compose file to the git repo
and have it build the Linux binary itself.

Note that it uses a somewhat “inefficient” builder step, i.e. not
combining steps together to reduce layers, as it uses a multi-stage
build to have a lean final image. This makes it easier to re-build if
something needs to change, as the cache is used more optimally.

For example, if only some go files change, most of the build is already
cached and only the go step gets re-run.

The more “efficient” builder step would look like this, but would have
to build the docs, web app and go CLI for any change in any file:

```Dockerfile
FROM golang:1.19-bullseye as builder

RUN apt-get update && \
    curl -fsSL https://deb.nodesource.com/setup_18.x | bash && \
    apt-get install -y \
    build-essential \
    nodejs \
    python3-pip

WORKDIR /app
ADD . .

RUN make web docs cli-linux-server
```

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,3 @@
+dist
+*/node_modules
+Dockerfile*
--- a/Dockerfile-build
+++ b/Dockerfile-build
@@ -1,0 +1,53 @@
+FROM golang:1.19-bullseye as builder
+
+ARG VERSION=dev
+ARG COMMIT=unknown
+
+RUN apt-get update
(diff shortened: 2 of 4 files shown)


binwiederhier:
Looks good. Thanks

PR #766, opened 2023-06-03
*: python pip in debian bookworm/testing needs venv

vbatts:
When testing this build using the upcoming debian bookworm/testing, the python pip v3.11 fails to install requirements that may clobber the host install packages. It prints out that venv must be used.

This change works fine on the current debian bullseye, and will continue to work once folks switch to the upcoming debian bookworm release.

### DIFF
--- a/Dockerfile-build
+++ b/Dockerfile-build
@@ -8,7 +8,8 @@ RUN curl -fsSL https://deb.nodesource.com/setup_18.x | bash
 RUN apt-get install -y \
     build-essential \
     nodejs \
-    python3-pip
+    python3-pip \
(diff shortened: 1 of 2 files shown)


binwiederhier:
Thanks for the PR.

There's another PR regarding Python, https://github.com/binwiederhier/ntfy/pull/665, I finally need to get a grip on that. It's been too long.

binwiederhier:
I re-did this in https://github.com/binwiederhier/ntfy/commit/c0fab933a5a4e144654c01d30318d5a3d1ec71b7. Sorry it took so long.

PR #774, opened 2023-06-17
Update Docker-Compose Version in install.md

skittlesvampir:
According to https://docs.docker.com/compose/compose-file/compose-file-v2/#healthcheck, 'start_period' is only supported since version 2.3


binwiederhier:
Thanks

PR #783, opened 2023-06-26
Add version info to web app

nimbleghost:
screenshot

### DIFF
--- a/Dockerfile-build
+++ b/Dockerfile-build
@@ -24,7 +24,7 @@ RUN make docs-build
 ADD ./web/package.json ./web/package-lock.json ./web/
 RUN make web-deps
 ADD ./web ./web
-RUN make web-build
+RUN make VERSION=$VERSION COMMIT=$COMMIT web-build
 
 # cli & server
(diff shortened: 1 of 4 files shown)


binwiederhier:
I am actually disagreeing with this one. Exposing the version of any running software on the web makes it easier to fingerprint it for vulnerabilities. That's why nginx et al have options to turn version signatures off. I'd be okay with showing it in dev mode only, which isn't all that useful.

nimbleghost:
I get your point, but it's fairly simple to determine via the bundler version hashing (if you have access to the releases to compare it to).

binwiederhier:
> I get your point, but it's fairly simple to determine via the bundler version hashing (if you have access to the releases to compare it to).

That's true, but let's not make it too easy for the script kiddies.

theS1LV3R:
How about having it be an option? I am using [argus](https://release-argus.io) for monitoring updates to the software I run. Having version information be exposed either over the API or in the GUI would make that monitoring way easier.

PR #795, opened 2023-06-29
PWA: Fix reload, Firefox mp3 load, reduce mobile padding

nimbleghost:


binwiederhier:
The "hotfix" for the .mp3 stuff only required the .mp3 change, right? This PR includes a bunch of things. Hotfixes should only ever include the minimal possible change to make what's released already work.

web/src/components/theme.js
```diff
@@ -55,6 +55,13 @@ export const darkTheme = {
...
       },
     },
+    MuiPaper: {
+      styleOverrides: {
+        root: {
+          backgroundImage: "none",
+        },
+      },
+    },
   },
```
binwiederhier, line 65: Can you explain this? Where do we have a background image that would get in the way?

web/src/index.jsx
```diff
@@ -1,6 +1,34 @@
...
+          cache: "no-store",
+          "cache-control": "no-cache",
+        },
+      });
+
+      if (resp?.status === 200) await registration.update();
+    }, intervalMS);
+  },
+});
+
```
binwiederhier, line 32: Can you move this to some other file, so that the index.jsx doesn't become awkward and hard to read?

web/vite.config.js
```diff
@@ -16,7 +16,7 @@ export default defineConfig(({ mode }) => ({
     react(),
     VitePWA({
       registerType: "autoUpdate",
-      injectRegister: "inline",
+      injectRegister: null,
```
binwiederhier, line 19: Can you add a comment `// Service worker is registered in XYZ`

binwiederhier (approved):
Good stuff. 

Normally I'd do the changes myself and roll with it, but I'm too tired. Turns out real work is exhausting.

web/src/components/theme.js
nimbleghost: will add a comment - there was an odd greyish background being added by MUI to the sidebar on narrow screens

| before | after |
| --- |  --- |
| before | after |

PR #804, opened 2023-07-03
Web app: add RTL support

### LINKED ISSUE #663: Support RTL layout
Hi,

The ntfy Web UI is not yet supporting RTL `RightToLeft`. May be it could be added later for example for arabic, farsi and hebrew.

For now, the layout is LTR `LeftToRight` only.

### LINKED ISSUE #700: Date/time not following system locale
I'm using polish locale, but date/time of messages in webui is american. I have 12h clock and MM/dd/YY.

Web server

image

Date/time format should follow system locale or be configurable. I'm using latest ntfy.

nimbleghost:
Resolves #663
Resolves #700

MUI has pretty good RTL support, not much to do except connect mui & i18next.

I do not read an RTL language, so probably want someone else to double check that it looks correct, but going by my basic knowledge and the Material reference, it seems fine.

https://mui.com/material-ui/guides/right-to-left
https://m2.material.io/design/usability/bidirectionality.html

---

| . | . | . |
| -- | -- | -- |
|  CleanShot 2023-07-03 at 15 23 13@2x | CleanShot 2023-07-03 at 15 16 36@2x | CleanShot 2023-07-03 at 15 14 37@2x |
|  CleanShot 2023-07-03 at 15 25 27@2x | CleanShot 2023-07-03 at 15 16 56@2x | CleanShot 2023-07-03 at 15 17 00@2x | 

CleanShot 2023-07-03 at 15 25 58@2x

### DIFF
--- a/web/src/components/App.jsx
+++ b/web/src/components/App.jsx
@@ -3,6 +3,7 @@ import { createContext, Suspense, useContext, useEffect, useState, useMemo } fro
 import { Box, Toolbar, CssBaseline, Backdrop, CircularProgress, useMediaQuery, ThemeProvider, createTheme } from "@mui/material";
 import { useLiveQuery } from "dexie-react-hooks";
 import { BrowserRouter, Outlet, Route, Routes, useParams } from "react-router-dom";
+import { useTranslation } from "react-i18next";
 import { AllSubscriptions, SingleSubscription } from "./Notifications";
 import { darkTheme, lightTheme } from "./theme";
 import Navigation from "./Navigation";
@@ -21,6 +22,7 @@ import Signup from "./Signup";
 import Account from "./Account";
 import "../app/i18n"; // Translations!
 import prefs, { THEME } from "../app/Prefs";
+import RTLCacheProvider from "./RTLCacheProvider";
 
 export const AccountContext = createContext(null);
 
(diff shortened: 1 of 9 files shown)


web/src/components/App.jsx
```diff
@@ -39,37 +41,47 @@ const darkModeEnabled = (prefersDarkMode, themePreference) => {
...
-    () => createTheme(darkModeEnabled(prefersDarkMode, themePreference) ? darkTheme : lightTheme),
-    [prefersDarkMode, themePreference]
+    () => createTheme({ ...(darkModeEnabled(prefersDarkMode, themePreference) ? darkTheme : lightTheme), direction: languageDir }),
+    [prefersDarkMode, themePreference, languageDir]
   );
 
+  useEffect(() => {
+    document.documentElement.setAttribute("lang", i18n.language);
+    document.dir = languageDir;
+  }, [i18n.language, languageDir]);
```
binwiederhier, line 59: Dumb question: 
- Should this not be `i18n.resolvedLanguage` everywhere? I had to do that in the preferences to avoid issues.
- Have you tested this with a cleared localStorage, i.e. without a language?

web/src/components/RTLCacheProvider.jsx
```diff
@@ -0,0 +1,22 @@
...
+  stylisPlugins: [prefixer, rtlPlugin],
+});
+
+const RTLCacheProvider = ({ children }) => {
+  const { i18n } = useTranslation();
+
+  return i18n.dir() === "rtl" ? <CacheProvider value={cacheRtl}>{children}</CacheProvider> : children;
+};
+
+export default RTLCacheProvider;
```
binwiederhier, line 22: I know the guide says "use a cache provider", but i couldn't figure out why. Do you happen to know the answer?

binwiederhier (approved):
This looks great! I had two questions but I'm ready to merge.

web/src/components/App.jsx
nimbleghost: - `resolvedLanguage` is specifically for preferences - it references a value in your supported languages, so you can match it to an option with the user. `language` is fine for all other use cases.

nimbleghost: - `resolvedLanguage` is specifically for preferences - it references a value in your supported languages, so you can match it to an option with the user. `language` is fine for all other use cases: https://www.i18next.com/overview/api#language
- works fine in incognito. it will use the fallback language if it can't detect one based on what the browser reports, or if it doesn't match a supported language.

web/src/components/RTLCacheProvider.jsx
nimbleghost: It's for caching the results of the CSS plugins (vendor autoprefixing, generated class names). Theres's a default cache  too; this one adds on the RTL plugin to the default plugins and then uses a separate cache.

binwiederhier: No idea what that means, hehe, but I trust you. :-D

PR #811, opened 2023-07-05
Add new integration ansible-role-ntfy-alertmanager

bleetube:
I also wrote a [ansible-role-ntfy](https://github.com/bleetube/ansible-role-ntfy) a few months ago, but I noticed there is a much better role for installing ntfy on the list already.


binwiederhier:
Awesome. Thank you!

PR #815, opened 2023-07-08
Markdown

binwiederhier:


server/server_test.go
```diff
@@ -1518,6 +1518,39 @@ func TestServer_PublishActions_AndPoll(t *testing.T) {
 	require.Equal(t, "target_temp_f=65", m.Actions[1].Body)
 }
 
+func TestServer_PublishMarkdown(t *testing.T) {
+	s := newTestServer(t, newTestConfig(t))
+	response := request(t, s, "PUT", "/mytopic", "_underline this_", map[string]string{
```
wunter8, line 1523: All of these `_underline this_` are technically _italics_

server/server_test.go
```diff
@@ -1535,12 +1568,25 @@ func TestServer_PublishAsJSON(t *testing.T) {
 	require.Equal(t, "google.pdf", m.Attachment.Name)
 	require.Equal(t, "http://ntfy.sh", m.Click)
 	require.Equal(t, "https://ntfy.sh/static/img/ntfy.png", m.Icon)
+	require.Equal(t, "", m.ContentType)
```
nihalgonsalves, line 1571: since it's a new field, maybe we can have it set to `text/plain` when not markdown? makes it easier to do match against in the future if needed.

nihalgonsalves (approved):

server/server_test.go
binwiederhier: Fixed

server/server_test.go
binwiederhier: I don't see the need to. The default has always been plaintext, and without a content type that's assumed. 99% of messages are plaintext, and adding the content type would just be useless bytes everywhere.

PR #816, opened 2023-07-10
Refine user header code

binwiederhier:
Refined #812

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -1858,10 +1858,52 @@ func (s *Server) autorizeTopic(next handleFunc, perm user.Permission) handleFunc
 	}
 }
 
-// maybeAuthenticate reads the "Authorization" header and will try to authenticate the user
+// maybeAuthenticate delegates between auth based on the Authorization header (Bearer/Basic), and auth
+// based on the user-defined header (as defined in the "auth-user-header" setting). The function prefers
+// the user-defined header, if both are present.
+//
+// This function will ALWAYS return a visitor, even if an error occurs (e.g. unauthorized), so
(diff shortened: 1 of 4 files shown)


server/server.go
```diff
@@ -1858,24 +1858,61 @@ func (s *Server) autorizeTopic(next handleFunc, perm user.Permission) handleFunc
...
+// that subsequent logging calls still have a visitor context.
+func (s *Server) maybeAuthenticate(r *http.Request) (*visitor, error) {
+	ip := extractIPAddress(r, s.config.BehindProxy)
+	vip := s.visitor(ip, nil) // IP-based visitor
+	if s.userManager == nil {
+		return vip, nil
+	}
+	if s.config.AuthUserHeader != "" && s.config.BehindProxy {
+		username := r.Header.Get(s.config.AuthUserHeader) // Do not allow a query param, only a header!
+		if username != "" {
```
Nanowires, line 1875: Up to you, but shouldn't this be located inside authenticateViaUserDefinedHeader(), to keep the abstraction layer more consistent (i.e. for the auth header authn, the header is alos read in authenticateViaUserDefinedHeader(), not in the caller function)?

Nanowires (approved):
See comment above, but looks good, from my perspective :-)

Nanowires:
I've added the tests to the brnch of the original PR, together with your changes in #812

server/server.go
Nanowires: I did realize, that if the username is not provided, you will fall back to the default authorization, so ignore my comment.

PR #817, opened 2023-07-10
fix(web-push): re-init i18n on each sw message

nihalgonsalves:
It looks like the context that Safari runs the service worker in does not have the global i18n initialisation. This changes the push handler to get a new i18n TFunction when handling each message. 

Before:

iOS screenshot with a notification showing i18n key instead of actual text

After:

iOS screenshot with a notification showing actual translated text

### DIFF
--- a/Dockerfile-build
+++ b/Dockerfile-build
@@ -1,4 +1,4 @@
-FROM golang:1.19-bullseye as builder
+FROM golang:1.20-bullseye as builder
 
 ARG VERSION=dev
 ARG COMMIT=unknown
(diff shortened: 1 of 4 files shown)


binwiederhier:
This incidentally has happened to me as well. Good stuff.

binwiederhier:
Wait, will this request the en.json file on every message? I mean it's not terrible since it's cached, but still...

nihalgonsalves:
@binwiederhier not every message - on every unknown type or expiring message, plus when you click a notification.

It is cached by the service worker itself so it is indeed fine, but I should have also moved the i18n init call for the click action to inside the action `catch` block..

PR #819, opened 2023-07-12
Tweak httpd config to use less resources

nisbet-hubbard:
Many thanks for writing this great piece of software and making it open-source! Being able to self-host `ntfy` really goes a long way towards removing one of the weak links in the Matrix ecosystem, which is the reliance on a central push server that can easily get blocked in certain parts of the world.

Just a couple of tweaks mostly to the `httpd` configurations to substitute the deprecated `mod_proxy_wstunnel`, and use more efficient alternatives to `mod_rewrite` mainly for the benefit of constrained environments.

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -44,6 +44,14 @@ Here are a few working sample configs:
     attachment-cache-dir: "/var/cache/ntfy/attachments"
     ```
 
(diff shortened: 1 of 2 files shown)


binwiederhier:
Thank you for this!

PR #821, opened 2023-07-14
DOC: Update grafana address in config.md

Albirew:
Grafana picture URI and URL were wrong. updated them to reflect reality on Github


binwiederhier:
The link is actually correct when compiled with `mkdocs`. Thank your for your PR though!

PR #835, opened 2023-08-08
Extra fields in JSON body

### LINKED ISSUE #827: Allow sending custom properties in POSTed JSON
:bulb: **Idea**
When publishing a notification using JSON, it would be useful to allow sending additional custom properties which will be forwarded verbatim to the mobile front end (assuming one uses a self-hosted ntfy server and builds their own mobile app). These additional properties could be nested under the toplevel `"extras"` property in the POSTed JSON.

This is a feature request for the ntfy server.

Example:

```
{
  "topic": "1234", 
  // existing properties
  "message": "foo",
  "title": "bar",
  // additional custom properties
  "extras": {
    "customField": "baz"
  }
}
```

wunter8:
fixes #827

### DIFF
--- a/docs/publish.md
+++ b/docs/publish.md
@@ -1114,6 +1114,7 @@ all the supported fields:
 | `delay`    | -        | *string*                         | `30min`, `9am`                            | Timestamp or duration for delayed delivery                            |
 | `email`    | -        | *e-mail address*                 | `phil@example.com`                        | E-mail address for e-mail notifications                               |
 | `call`     | -        | *phone number or 'yes'*          | `+1222334444` or `yes`                    | Phone number to use for [voice call](#phone-calls)                    |
+| `extras`   | -        | *JSON object*                    | `{"customField": "customValue"}`          | Extra key:value pairs will be included in the notification            |
 
 ## Action buttons
 _Supported on:_ :material-android: :material-apple: :material-firefox:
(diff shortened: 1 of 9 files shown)


server/server.go
```diff
@@ -1010,6 +1010,14 @@ func (s *Server) parsePublishParams(r *http.Request, m *message) (cache bool, fi
 			return false, false, "", "", false, errHTTPBadRequestActionsInvalid.Wrap(e.Error())
 		}
 	}
+	extrasStr := readParam(r, "x-extras")
```
binwiederhier, line 1013: Probably add `extras` to this list too

docs/publish.md
```diff
@@ -2661,6 +2662,123 @@ Here's an example of how it will look on Android:
...
+    ``` php-inline
+    file_get_contents('https://ntfy.sh/mytopic', false, stream_context_create([
+        'http' => [
+        'method' => 'PUT',
+        'header' =>
+            "Content-Type: text/plain\r\n" . // Does not matter
+            "Title: Low disk space alert\r\n" .
+            "Tags: warning,cd\r\n" .
+            "X-Priority: 4\r\n" .
+            "X-Extras: {\"lastChecked\": \"20230205\"}",
```
binwiederhier, line 2776: Welllllll, that is just really shitty syntax, no? I feel like we should also add the "simple" syntax we have for the actions here. We could probably re-use the parser for most of it, though it's suddenly a lot harder.

binwiederhier (changes requested):
Well done! 

There are a few things missing:

- Persistence. If we want extras to be a first class citizen, we need to store it in the database. Otherwise they won't show up when polling.
- Are you doing the Android side of this too :grimacing: 
- The "simple" parsing would be nice, though I'd probably be okay doing this in a second pass

wunter8:
> * Persistence. If we want extras to be a first class citizen, we need to store it in the database. Otherwise they won't show up when polling.
You're right. I totally forgot about this. I was planning on doing a schema change on the Android side, but I forgot about the server DB.

> * Are you doing the Android side of this too 😬
Yes, I was planning on it. I'm hoping nothing has broken since the last time I looked at the Android side of things

> * The "simple" parsing would be nice, though I'd probably be okay doing this in a second pass
I'll have to remind myself how the simple parsing works for Actions. I think it'd have to be different, though, because here we have a list of key:value pairs that can be anything. The Actions have a few possible keys to choose from and a limited number of keys total.

server/server.go
wunter8: 👍

I noticed in the docs that `extras` are already an option in Intent actions. Are you okay with having that term used here and there?

docs/publish.md
wunter8: I didn't test most of these and don't know if the syntax is actually correct or if there's an easier way to write it out, especially with this PHP one. I'll look into adding the "simple" syntax, though

binwiederhier:
> I'll have to remind myself how the simple parsing works for Actions. I think it'd have to be different, though, because here we have a list of key:value pairs that can be anything. The Actions have a few possible keys to choose from and a limited number of keys total.

The parsing thing first parses key value pairs inside each section afaik, so it should fit nicely. There's also the arbitrary header parsing in there for `headers.Content-Type=...` or something like that

binwiederhier:
Any update on this?

wunter8:
I looked at the "simple" parsing code, and it's anything but simple 😂 I think I'm going to hold off on that for now. It seems like it would require a decent sized refactor

tka85:
Would be pretty useful to have the `extras` field for schema-less JSON payloads.

Any chance this PR could get some love?

server/message_cache.go
```diff
@@ -700,6 +717,12 @@ func readMessage(rows *sql.Rows) (*message, error) {
 			return nil, err
 		}
 	}
+	var extras map[string]string
```
goodevilgenius, line 720: What if, instead of a `map[string]string`, you used a `map[string]json.RawMessage`?

That way, you should be able to use any arbitrary JSON, instead of just strings.

https://play.golang.com/p/AZ2XzbdqnXk

camerow:
+1 to this being helpful!

tka85:
Closed because it was implemented or because it is no considered meaningful/of interest any longer?

PR #842, opened 2023-08-18
Make UP return 410 after topic has expired

### LINKED ISSUE #664: UnifiedPush: Respond with 404/409/... to Mastodon/etc. instead of 507 based on User-Agent
Similar to the Matrix Push Gateway change, we should respond to Mastodon/etc. in a way that will make them delete the push subscription. 

Mastodon:
https://github.com/mastodon/mastodon/blob/730bb3e211a84a2f30e3e2bbeae3f77149824a68/app/workers/web/push_notification_worker.rb#L35-L46

Common-Proxies:
https://github.com/UnifiedPush/common-proxies/blob/main/gateway/matrix.go#L63

/cc FYI @karmanyaahm @p1gp1g

ShadowJonathan:
This resolves https://github.com/binwiederhier/ntfy/issues/664;

If a subscription topic;
- Does not have its `t.rateVisitor.Stale()`
- Does not have any subscribers;
- And the time since last access (including poll and create) has been 16h

...it will start responding with 410s.

This is to make sure that topics eventually will get deleted on servers like mastodon.

There should not be any race condition on first create; Create sets `lastAccess` to `now()` the same as a Poll would, through `topicsFromIDs`, which gets indirectly called from http middleware every time a topic is mentioned, such as [here](https://github.com/ShadowJonathan/ntfy/blob/dd02267f9b3e314ed04948c24da172477af9d26a/server/server_middleware.go#L31)

Unless a client somehow reuses a topic, and sends a topic publish link to the application, which uses it *before* the client has started its first poll on the topic, this should not create a race condition.

### DIFF
--- a/server/errors.go
+++ b/server/errors.go
@@ -125,6 +125,7 @@ var (
 	errHTTPConflictSubscriptionExists                = &errHTTP{40903, http.StatusConflict, "conflict: topic subscription already exists", "", nil}
 	errHTTPConflictPhoneNumberExists                 = &errHTTP{40904, http.StatusConflict, "conflict: phone number already exists", "", nil}
 	errHTTPGonePhoneVerificationExpired              = &errHTTP{41001, http.StatusGone, "phone number verification expired or does not exist", "", nil}
+	errHTTPGoneSubscriptionExpired                   = &errHTTP{41002, http.StatusGone, "subscription expired", "", nil}
 	errHTTPEntityTooLargeAttachment                  = &errHTTP{41301, http.StatusRequestEntityTooLarge, "attachment too large, or bandwidth limit reached", "https://ntfy.sh/docs/publish/#limitations", nil}
 	errHTTPEntityTooLargeMatrixRequest               = &errHTTP{41302, http.StatusRequestEntityTooLarge, "Matrix request is larger than the max allowed length", "", nil}
 	errHTTPEntityTooLargeJSONBody                    = &errHTTP{41303, http.StatusRequestEntityTooLarge, "JSON body too large", "", nil}
--- a/server/server.go
+++ b/server/server.go
@@ -742,7 +742,14 @@ func (s *Server) handlePublishInternal(r *http.Request, v *visitor) (*message, e
 	if e != nil {
 		return nil, e.With(t)
 	}
-	if unifiedpush && s.config.VisitorSubscriberRateLimiting && t.RateVisitor() == nil {
+	if unifiedpush && s.config.VisitorSubscriberRateLimiting && t.Stale() {
+		// This subscription is verifiably dead, we should notify the server.
+		//
+		// This not cause the below condition (where ratevisitor is nil) to be excluded,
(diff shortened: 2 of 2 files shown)


ShadowJonathan:
Actually, i just double-checked, and this approach will not work, because the Server Manager will prune these topics every minute or so.

Then, when they are created again, their timers are reset.

Even if I were to up this expunge time from 16h to 24h, and/or add an extra test for "if the topic is decaying" (12h), I still think that can be a bit of a broken approach, since it'll basically make a dead topic "removable" only 4h after 12h of retries after it has been referenced by the server again.

ShadowJonathan:
I'd like to get some feedback on the direction of this change before I'll poke to fix coverage

karmanyaahm:
> since it'll basically make a dead topic "removable" only 4h after 12h of retries after it has been referenced by the server again.

Yeah. You're right, this is just delaying the inevitable by a (relatively small) margin. If this really needs immediate attention, just change the errHTTPInsufficientStorageUnifiedPush to 410 gone.

Like I said before, we'll have to modify the Android app to send NEW_ENDPOINT on reconnect, and then changing errHTTPInsufficientStorageUnifiedPush to 410 gone will be harmless. I could *maybe* attempt that over the next week, but no guarantees, it's my first week of college. If you can do it, that would be epic.

ShadowJonathan:
I'm more concerned how ntfy integrated in other apps would react to something like this, we see this from a mastodon app, most likely fedilab, which means they're using this API for something, what do you think about that?

binwiederhier:
@karmanyaahm @ShadowJonathan I have no personal stake in this, so I defer to your judgement of whether I should merge this or not. It may be worth thinking through possible race conditions and such. Telling a client to delete the subscription is a serious thing. If we are wrong, a lot of people will start complaining about not receiving notifications anymore.

karmanyaahm:
Yeah, let's merge this. Just give me this week to attempt Android re-registration on expiry modifications. It's thanksgiving break so I'm hoping to seriously really get this solved, it's been too long.

ShadowJonathan:
I'm not sure how to deal with the codecov runs, unfortunately

karmanyaahm:
Doing this in #979, because it requires some other changes to avoid losing subscriptions permanently.

server/errors.go
```diff
@@ -125,6 +125,7 @@ var (
 	errHTTPConflictSubscriptionExists                = &errHTTP{40903, http.StatusConflict, "conflict: topic subscription already exists", "", nil}
 	errHTTPConflictPhoneNumberExists                 = &errHTTP{40904, http.StatusConflict, "conflict: phone number already exists", "", nil}
 	errHTTPGonePhoneVerificationExpired              = &errHTTP{41001, http.StatusGone, "phone number verification expired or does not exist", "", nil}
+	errHTTPGoneSubscriptionExpired                   = &errHTTP{41002, http.StatusGone, "subscription expired", "", nil}
```
ShadowJonathan, line 128:
```suggestion
	errHTTPGoneSubscriptionExpired                   = &errHTTP{41002, http.StatusGone, "subscription has expired", "", nil}
```

Issue #461, opened 2022-11-01
iOS App not receiving any notifations with 'Priority' set


pinpox:
Not sure if it's a duplicate of #353, but the iOS app fails to receive any notications that have `Priority` set.
Example:
This is not received on my mobile phone
```
curl -u $NTFY_USER:$NTFY_PASS \                                                                                                                                 impure ❄ ntfy-sh
  -H 'Title: Backup on ahorn failed!' \
  -H 'Priority: high' \
  -H 'Tags: backup,borg,ahorn' \
  -d 'Backup error on ahorn: $exitStatus' 'https://push.pablo.tools/pinpox_backups'
  ```
While this works as expected:
```
curl -u $NTFY_USER:$NTFY_PASS \                                                                                                                                 impure ❄ ntfy-sh
  -H 'Title: Backup on ahorn failed!' \
  -H 'Tags: backup,borg,ahorn' \
  -d 'Backup error on ahorn: $exitStatus' 'https://push.pablo.tools/pinpox_backups'
```

I've tried different Priority values, they all seem to break the delivery on iOS. Subscribing to the topic via the CLI shows everything fine, it seems to be related to iOS.

**Comment by binwiederhier** [comment]
1. iOS is quite buggy, so I don't doubt that there could be issues. That said, the iOS priority is not adjusted in the Firebase message, so it would have to be on the iOS side.
2. Are you behind Cloudflare? There is this ticket #351 which could be related.
3. Can you post the server logs with `log-level: TRACE` set? Just to rule out that this is a server-side issue?
4. You could probably censor your hostname, unless you're ok with people knowing it.

**Comment by 0x111** [comment]
I can confirm that this is happening in my case too. The desktop version displays both, while the ios app only the one without the priority flag. 

`ntfy version: ntfy 1.28.0 (d3cfa34), runtime go1.18.6, built at 2022-09-27T16:55:57Z`

I did remove ip addresses and instance urls, but this is the Trace logs. I was sending a ntfy message from the same server where ntfy is running momentarily. I also used the same messages as provided by the OP.

```
2022/11/13 09:59:10 INFO Listening on 127.0.0.1:4456[http], ntfy 1.28.0, log level is TRACE
2022/11/13 09:59:15 DEBUG server-ipv6 HTTP POST /argus-uptime Dispatching request
2022/11/13 09:59:15 TRACE server-ipv6 HTTP POST /argus-uptime Entire request (headers and body):
POST /argus-uptime HTTP/1.1
[...]

Issue #353, opened 2022-07-01
Bug: iOS client fails to receive messages due to Cloudflare injecting invalid priority header


ksurl:
log:

```
2022/06/30 17:00:13 DEBUG 2607:fb91:2a2:e023:508e:6746:ebcd:3f4f HTTP GET /APP/json?poll=1&since=vN1R3yK4zHhO Dispatching request
2022/06/30 17:00:13 TRACE 2607:fb91:2a2:e023:508e:6746:ebcd:3f4f HTTP GET /APP/json?poll=1&since=vN1R3yK4zHhO Entire request (headers and body):
GET /APP/json?poll=1&since=vN1R3yK4zHhO HTTP/1.1
Accept: */*
Cdn-Loop: cloudflare
Cf-Ray: 723adc32fc02523f-LAX
Cf-Visitor: {"scheme":"https"}
Accept-Encoding: gzip
Accept-Language: en-US,en;q=0.9
Cf-Ipcountry: US
X-Forwarded-For: 2607:fb91:2a2:e023:508e:6746:ebcd:3f4f, 2607:fb91:2a2:e023:508e:6746:ebcd:3f4f
X-Forwarded-Host: ntfy.DOMAIN
X-Forwarded-Proto: https
X-Forwarded-Ssl: on
Authorization: Basic AUTH
Cf-Connecting-Ip: 2607:fb91:2a2:e023:508e:6746:ebcd:3f4f
Connection: close
X-Real-Ip: 2607:fb91:2a2:e023:508e:6746:ebcd:3f4f
User-Agent: ntfy/1.2 (build 3; iOS 15.5.0)
Cf-Warp-Tag-Id: f6e8ef16-b59e-49c4-bab3-7f8a619ab082
Priority: u=3
2022/06/30 17:00:13 DEBUG 2607:fb91:2a2:e023:508e:6746:ebcd:3f4f HTTP GET /APP2/json?poll=1&since=all Dispatching request
2022/06/30 17:00:13 TRACE 2607:fb91:2a2:e023:508e:6746:ebcd:3f4f HTTP GET /APP2/json?poll=1&since=all Entire request (headers and body):
GET /APP2/json?poll=1&since=all HTTP/1.1
Authorization: Basic AUTH
Cdn-Loop: cloudflare
Accept-Language: en-US,en;q=0.9
Cf-Ipcountry: US
Cf-Ray: 723adc3308ff7ad8-LAX
Cf-Visitor: {"scheme":"https"}
[...]

Issue #351, opened 2022-07-01
Ignore new HTTP Priority header


binwiederhier:
Cloudflare apparently adds a `Priority: u=...` header to requests, which ntfy rejects with `500 invalid priority` (should be a 400). This is a new HTTP standard (https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-priority) that is emerging.

We should ignore header values that start with `u=`. It is quite unfortunate that the name conflicts like this.

binwiederhier:
This ticket (https://github.com/binwiederhier/ntfy/issues/353) has a full dump of the Cloudflare request showing the `Priority: u=3` header.

PR #851, opened 2023-08-29
Ignore Cloudflare Priority header

gusdleon:
# Refine handling of HTTP Priority header

Part of: #351, #353, #461, ...

This PR improves the handling of the new HTTP Priority header (RFC 9218) commonly affecting ntfy instances running behind Cloudflare.

## Changes:
- Now if the received PUT/POST request contains the mentioned header simply ignores it and continues searching for another header or query parameter
- This PR reverts some changes in 95bd876

A working instance with this fix can be tested [Here](https://ntfy.gusdleon.com/)
😀

### DIFF
--- a/server/util.go
+++ b/server/util.go
@@ -9,6 +9,7 @@ import (
 	"net/http"
 	"net/netip"
 	"strings"
+	/*"regexp"*/
 )
 
 var mimeDecoder mime.WordDecoder
@@ -129,7 +130,25 @@ func fromContext[T any](r *http.Request, key contextKey) (T, error) {
 func maybeDecodeHeader(header string) string {
 	decoded, err := mimeDecoder.DecodeHeader(header)
 	if err != nil {
-		return header
+		return cloudflarePriorityIgnore(header)
 	}
-	return decoded
+	return cloudflarePriorityIgnore(decoded)
+}
+
+// Ignore new HTTP Priority header (see https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-priority)
+// Cloudflare adds this to requests when forwarding to the backend (ntfy), so we just ignore it.
+// If the Priority header is set to "u=*, i" or "u=*" (by cloudflare), the header will be ignored.
+// And continue searching for another header (x-priority, prio, p) or in the Query parameters.
+func cloudflarePriorityIgnore(value string) string {
+	if strings.HasPrefix(value, "u=") {
+		return ""
+	}
+
+	// The same but with regex
+	/* pattern := `^u=\d+\s*,\s*i|u=\d+$`
+	regex := regexp.MustCompile(pattern)
+	if regex.MatchString(value) {
+		return ""
+	} */
+	return value
 }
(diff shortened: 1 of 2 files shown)


danricho:
+1 for this as it would enable me to use NTFY with Uptime Kuma behind cloudflare. It inherently sets the priority of alerts which don’t come through on iOS properly.

server/util.go
```diff
@@ -9,6 +9,7 @@ import (
 	"net/http"
 	"net/netip"
 	"strings"
+	/*"regexp"*/
```
binwiederhier, line 12: ^

server/util.go
```diff
@@ -129,7 +130,25 @@ func fromContext[T any](r *http.Request, key contextKey) (T, error) {
...
+		return ""
+	}
+
+	// The same but with regex
+	/* pattern := `^u=\d+\s*,\s*i|u=\d+$`
+	regex := regexp.MustCompile(pattern)
+	if regex.MatchString(value) {
+		return ""
+	} */
+	return value
```
binwiederhier, line 153: 1. This is commented code
2. The func comment doesn't match with the implementation
3. Go function comments start with the function name, e.g. `cloudflarePriorityIgnore ignores the ...`

server/util.go
```diff
@@ -129,7 +130,25 @@ func fromContext[T any](r *http.Request, key contextKey) (T, error) {
 func maybeDecodeHeader(header string) string {
 	decoded, err := mimeDecoder.DecodeHeader(header)
 	if err != nil {
-		return header
+		return cloudflarePriorityIgnore(header)
 	}
-	return decoded
+	return cloudflarePriorityIgnore(decoded)
```
binwiederhier, line 135: This function is used not just by the priority, but also by other headers. So one could not use a notification title `u=hello` anymore. Why not keep this logic only in the parsePrio function like it was before?

binwiederhier (changes requested):

server/util.go
gusdleon: Sorry sorry sorry 🤯
Sorry for the error, I was so focussed trying to solve specifically the problem with the CF priority header that I didn't think about the possibility of another header having a value starting with `u=`, even knowing this function is used with with all headers.

> Why not keep this logic only in the parsePrio function like it was before? 

You mean this?
https://github.com/binwiederhier/ntfy/blob/2305ebca244fb60aa784a3b79270ae6f33b8ee48/util/util.go#L163-L170

Well, if you're referring to that part of the code, it's because at that point of execution, we have already parsed the headers and query parameters. If we find any priority header (priority, x-priority, prio, p), the others will not even be read, including the query parameters.

For instance, if a user is using the web interface to send a notification with priority 5, the request will include a query string like this:
`?title=titulodeprueba&priority=5`

If we send the request directly to the server, all will be ok, it will work, but if the server is behind CF, it will (in the majority of cases) have the `priority:"u=*"` header, causing the priority set in the query parameter to be ignored.

Nevertheless, I am fixing this issue in my next commit.
Thank you for all your comments.
😉

binwiederhier:
Refined and merged in https://github.com/binwiederhier/ntfy/pull/895 - Sorry it took so long. Thank you for fixing this!

PR #899, opened 2023-09-27
fix(pwa): hide install prompt on macOS 14 safari

nihalgonsalves:
macOS 14 has support for "Add to Dock", which added the `window.navigator.standalone` property to macOS. This was previously only on iOS and we relied on it to display the "install required" prompt.

Since the limitation of needing an installed web app does not exist on _desktop_ Safari, ntfy therefore wrongly stops you from using the web app notifications. This changes the heuristic to only display the install prompt when the push manager API is not available.

I also added a screenshot of the new Safari "Add to Dock" action to the docs.

While testing it I saw a deprecation warning when using the node source script, so I fixed that too - I can move that to another PR if wanted.

| Before | After |
| --- | --- |
| Screenshot showing the iOS alert wrongly appearing on macOS 14 | Fixed version, without the erroneous alert |


binwiederhier:
:+1:

PR #903, opened 2023-09-30
docs: ios shortcut

ohare93:
I made an iOS shortcut to share URLs between machines. It makes them clickable to open the url 👍 I saw there was only a fileshare iOS Shortcut in the docs (which is also done by executing curl using another app!) so I thought I'd share this one, which is much simpler and also does URLs.


ohare93:
Oops, I let prettier fix all the spacings on this docs page.  (Un)Fixed.

Issue #583, opened 2023-01-16
WebExtension


bt90:
It would be nice to have a basic WebExtension for ntfy similar to what Pushbullet is offering.

binwiederhier:
Nice idea. I do not think I have the capacity or time to make one, but I'm happy to accept contributions.

PR #906, opened 2023-10-01
Added browser add-on

jonathan-stein10:
Implementing suggestion #583.
The add-on is a very simple browser popup that displays the web app as an `iframe`. I've tested the add-on in Chrome and Firefox and would be happy to publish it in the respective stores.


binwiederhier:
Awesome. Thanks. I'll likely submit myself. It doesn't look like it needs to be updated often

binwiederhier:
I tested locally in Firefox by going to `about:debugging#/runtime/this-firefox` and clicking "Load temporary extension", and it seems that the notifications do not work. I even added `"permissions": ["notifications"]` to the `manifest.json` and it still showed the same warning. Notifications do not show as popup (Ubuntu, Firefox).

image

Issue #914, opened 2023-10-11
"Private instance" and ACL


sandman7920:
:lady_beetle: **Describe the bug**
If we have `auth-default-access: "deny-all"` and ACL for user `test test* rw`  and `test * ro`
 It seems like the `read-only access to topic * permission` is being applied before the read-write `access to topic test*`

:computer: **Components impacted**
ACL publish API

:bulb: **Screenshots and/or logs**
![image](https://github.com/binwiederhier/ntfy/assets/3589417/d1b997d2-6308-4691-817b-835211918276)

:crystal_ball: **Additional context**
![image](https://github.com/binwiederhier/ntfy/assets/3589417/20aad2d7-e042-4bd1-835c-f09def2ac225)

binwiederhier:
Fixed in #917. Thank you!

PR #917, opened 2023-10-13
user.Manager: fix ACL write, read order

sandman7920:
This should fix `read-only access to topic *` being applied before `read-write access to topic _PREFIX_*` Before this if we have:

```
ntfy access user "mytopic*" rw
ntfy access user "*" ro
```

read-only access rule was applied first and user couldn't write to mytopic*

See #914


sandman7920:
This PR still does not solve the following case
```
ntfy access user "mytopic*" rw
ntfy access user "mytopic_read_only*" ro
ntfy access user "*" ro
```

The user will still be able to write into `mytopic_read_onlyXXX`

Maybe query should be something like that
```go
selectTopicPermsQuery = `
  SELECT read, write
  FROM user_access a
  JOIN user u ON u.id = a.user_id
  WHERE (u.user = ? OR u.user = ?) AND ? LIKE a.topic ESCAPE '\'
  ORDER BY u.user DESC, LENGTH(a.topic) DESC, a.write DESC
`
```

For each user, we should test in order `THE_LONGEST_RULE`->`WRITE_PERMISSION`

binwiederhier:
Can you add a bunch of unit tests to check these cases? I don't want to just yolo merge this.

sandman7920:
I don't know how fast the `SQLIte`.`LENGTH`  function is, maybe it's reasonable to create a `topic_length` field in `access` table.

binwiederhier:
I would like to apologize for not addressing this earlier. I am looking at your solution now, and I'm trying to understand the root cause. Thank you for the tests and all the work. It is much appreciated!!

PR #919, opened 2023-10-18
doc/integrations: link RouterOS Scripts

eworm-de:
... which has a module to send notifications to Ntfy.


eworm-de:
Also available at Github... Would you prefer to link there?
https://github.com/eworm-de/routeros-scripts#routeros-scripts

binwiederhier:
Thank you!

PR #934, opened 2023-10-28
Fix #933 macos client location

### LINKED ISSUE #933: Add explicit client.yml default location for macOS install
ntfy client, macOS

:bulb: **Idea**
The current documentation lists the default client configuration file as being located either in 
  `~/.config/ntfy/client.yml` (for the non-root user)
or
  `/etc/ntfy/client.yml` (for the root user)

and on macOS you can create a `~/.config/ntfy/client.yml` file which is used by some linux based tools, but that is *not* where ntfy looks as described further down in the output of `ntfy pub --help`

The docs should be updated to be more explicit in noting the OS differences in the "From the CLI" portion and the "Installing ntfy" steps

bear:
While the location of the `client.yml` file for macOS is shown in docs for macOS https://github.com/binwiederhier/ntfy/blob/main/docs/install.md#macos it is not called out in the General Steps section or in the `subscribe/cli.md` part

To be thorough I've added the macOS file location to the other two locations

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -14,7 +14,7 @@ We support amd64, armv7 and arm64.
 
 1. Install ntfy using one of the methods described below
 2. Then (optionally) edit `/etc/ntfy/server.yml` for the server (Linux only, see [configuration](config.md) or [sample server.yml](https://github.com/binwiederhier/ntfy/blob/main/server/server.yml))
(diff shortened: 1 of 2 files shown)


binwiederhier:
Thank you!

PR #939, opened 2023-11-07
Fix broken badge image links in docs

artemislena:
T.: `/docs/` uses incorrect relative paths for the app store badges currently ('t worked on docs.ntfy.sh cause there `/docs/` 's `/`, but on GitHub n instances other than ntfy.sh it's broken); this should fix that. (I believe the issue occurred cause the paths might've been copy-pasted from `/docs/subscribe/phone/`?)


binwiederhier:
Thank you!

PR #949, opened 2023-11-17
Allow overriding `python3` & `pip3` binary in Makefile

cyqsimon:
This is a simpler (and hopefully better) attempt to solve the problem in #665.

I believe it is still needed because on some platforms like RHEL8 (which won't be EOL until 2029), `python3` still points to `python3.6` by default, which is too old for `mkdocs`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -120,7 +120,7 @@ docs-venv: .PHONY
 	$(PYTHON) -m venv ./venv
 
 docs-build: docs-venv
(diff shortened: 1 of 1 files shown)


binwiederhier:
:+1:

binwiederhier:
Not entirely sure why, but the build is failing with this: https://github.com/binwiederhier/ntfy/actions/runs/6899040178/job/18769998509?pr=949

cyqsimon:
Ah oops sorry my bad.

PR #961, opened 2023-11-24
Update docker-dev go version

cmj2002:
Running `go mod tidy` in go 1.21 will add `toolchain go<version>` to `go.mod` file, which cannot be recognized by previous go versions and will cause error. ([reference](https://github.com/golang/go/issues/62409)).

This PR updates the go version in `Dockerfile-build` to 1.21 to fix build errors.


binwiederhier:
Oops. Thanks!

Issue #572, opened 2023-01-06
Paste images from clipboard


rounakdatta:
Currently NTFY supports attaching a file from a local source or via an external URL. It would be cool if one could paste from clipboard and the file uploading just worked. An use case I can think of is quickly sharing an image from one computer to another.

binwiederhier:
On Android, if you click "Share" from any app, you can share an image or a file to a ntfy topic. There is no native dialog inside in the ntfy app itself yet (see #98) though.

On the web app, you can just drag and drop files into the web app and publish files like that.

rounakdatta:
Thanks for getting back. Here's a situation that prompted me to create this
issue: I use Flameshot on my Linux computer to take a selected-area
screenshot and instead of saving it to a file, I copy it to clipboard.

I think it'd be great if I could just Ctrl+V instead.

(The reason I copy to clipboard or even that am willing to share over NTFY
is that the screenshot is ephemeral and quick)

binwiederhier:
So you are talking about the web app?

binwiederhier:
Done in #963.

PR #963, opened 2023-11-25
Paste images from clipboard

cmj2002:
Implement #572.

Specific behaviors:
- When the clipboard contains an image and is pasted on the MessageBar, open the PublishDialog and attach the image in the clipboard as a local file.
- Clear the attached local file when the PublishDialog is closed.
- ~~Pasting inside the PublishDialog is ineffective.~~(edit: feature added)
- When the clipboard contains an image and is pasted on the the message section of PublishDialog, the attached file will be changed to the pasted image (no matter whether it already has a file or not)

For convenience in previewing, a new feature has also been added: clicking on the icon of the attached local file will create a blob URL using `URL.createObjectURL` and open it in a new tab. This feature is effective for all attached local files.

This feature has already been tested in Chrome and Firefox (on Windows). **Testing on Safari is still needed** (I don't have Apple devices). You can test this feature at https://ntfy-clipboard.fly.dev.

### DIFF
--- a/web/src/components/AttachmentIcon.jsx
+++ b/web/src/components/AttachmentIcon.jsx
@@ -1,5 +1,5 @@
 import * as React from "react";
-import { Box } from "@mui/material";
+import { Box, Link } from "@mui/material";
 import { useTranslation } from "react-i18next";
 import fileDocument from "../img/file-document.svg";
 import fileImage from "../img/file-image.svg";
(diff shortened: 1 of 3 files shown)


wunter8:
Hey! Thanks for adding this! It's been on my list of things to do. Maybe remove the two `console.log`s that you have in there?

Also, I think it might be nice to be able to paste an image inside the PublishDialog, inside the message area. Was there a specific reason you decided not to do that?

(binwiederhier will have the final say, of course)

cmj2002:
Thanks for your advises.

> Maybe remove the two console.logs that you have in there?

Sorry I forgot about this, this is what I use for debugging.

> Was there a specific reason you decided not to do that?

There is no specific reason, I will try to implement it later since you asked.

wunter8:
Looks great! Thanks again!

binwiederhier:
Thank you. Works amazingly.

PR #967, opened 2023-11-30
Added non-root user to docker images.

### LINKED ISSUE #966: Add non-root user to ntfy container images.
:bulb: **Idea**
I would like to run ntfy as a non-root user in my container. If we can add a non-root user to the container we would then be able to run ntfy as a standard user without having to build my own container based off of the ntfy official container.

ntfy server container

arahja:
Added non-root user to docker images. 
This gives you the ability to run ntfy as a non-root user. resolves #966


binwiederhier:
Thanks

binwiederhier:
I couldn't get this to build so I had to revert it. https://github.com/binwiederhier/ntfy/actions/runs/8195110634/job/22412872174

Feel free to open another PR with a fix for this. Try building via `make release-snapshot` to see if it would build o Github

arahja:
Sounds good, I will rework this PR.

arahja:
Ran the `make release-snapshot` build command as suggested. I was able to reproduce locally. I worked around the issue by using the full path for the adduser command. There seems to be a slight variance between the linux/arm/v7/alpine and the linux/amd64/alpine images related to the "$PATH" variable.

PR #979, opened 2023-12-16
Inform the client if it subscribes to a new topic

### LINKED ISSUE #664: UnifiedPush: Respond with 404/409/... to Mastodon/etc. instead of 507 based on User-Agent

karmanyaahm:
- Reject Matrix with 200 and rejected pushkey instead of returning 5xx
- Set UnifiedPush 'can't publish' error to 404
- Implement a NeverSubscribed() method in topics to keep track of new topics
- add the "new_topic" param to the open message
- Return new_topic in the open message when any of the list of topics subscribed to is new. Not the most information, but the likelihood of topics from the same client have different statuses is so tiny and rare, that clients can resubscribe all to be safe.

**Deployment**
I think the Android app needs to be released first, before the server is deployed, so the client can deal with the case of re-registration on 404.

Paired with https://github.com/binwiederhier/ntfy-android/pull/70
Closes https://github.com/binwiederhier/ntfy/issues/664

### DIFF
--- a/server/errors.go
+++ b/server/errors.go
@@ -117,9 +117,10 @@ var (
 	errHTTPBadRequestWebPushSubscriptionInvalid      = &errHTTP{40038, http.StatusBadRequest, "invalid request: web push payload malformed", "", nil}
 	errHTTPBadRequestWebPushEndpointUnknown          = &errHTTP{40039, http.StatusBadRequest, "invalid request: web push endpoint unknown", "", nil}
 	errHTTPBadRequestWebPushTopicCountTooHigh        = &errHTTP{40040, http.StatusBadRequest, "invalid request: too many web push topic subscriptions", "", nil}
-	errHTTPNotFound                                  = &errHTTP{40401, http.StatusNotFound, "page not found", "", nil}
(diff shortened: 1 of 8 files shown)


karmanyaahm:
Also, presumably the Android app needs to be pushed out first, before this server update is applied to give people time to update to the new_topic supporting system.

CrazyWolf13:
@binwiederhier 
Hi
Could you have a look at this PR yet, I would love to see this functionality, and seems like there is already a good PR made, which has now sadly been open and unreviewed for over a year.

Thanks for looking into this, once you find some time :)

Would really appreciate!

karmanyaahm:
Closing as this is 3 years old

PR #1002, opened 2024-01-15
Add systemd user service for `ntfy-client.service`

dandersch:
We can offer `ntfy-client.service` as a user service along the already existing system service.

This allows the use of `~/.config/ntfy/client.yaml` for configuration (for the system service, the config is hardcoded to use
`/etc/ntfy/client.yaml`). It also means that we can update the docs to recommend enabling the user service for "personal desktop machine" use in https://docs.ntfy.sh/subscribe/cli/?h=systemd#using-the-systemd-service and remove the part about using an `override.conf` (unless this has another use case for the ntfy-client that I am missing?).

Repos would need to update their packages to copy over `client/user/ntfy-client.service` to `/usr/lib/systemd/user/ntfy-client.service`.

I've been using it successfully so far. I've also added entries to `scripts/postinst.sh` and `.goreleaser.yml`, but haven't tested those.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -90,6 +90,8 @@ nfpms:
         type: "config|noreplace"
       - src: client/ntfy-client.service
         dst: /lib/systemd/system/ntfy-client.service
+      - src: client/user/ntfy-client.service
+        dst: /lib/systemd/user/ntfy-client.service
       - dst: /var/cache/ntfy
         type: dir
       - dst: /var/cache/ntfy/attachments
@@ -119,6 +121,7 @@ archives:
       - server/ntfy.service
       - client/client.yml
       - client/ntfy-client.service
+      - client/user/ntfy-client.service
   -
     id: ntfy_windows
     builds:
--- a/client/user/ntfy-client.service
+++ b/client/user/ntfy-client.service
@@ -1,0 +1,10 @@
+[Unit]
+Description=ntfy client
+After=network.target
+
+[Service]
+ExecStart=/usr/bin/ntfy subscribe --config "%h/.config/ntfy/client.yml" --from-config
+Restart=on-failure
+
+[Install]
+WantedBy=default.target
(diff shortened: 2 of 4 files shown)


Migs3:
Will this be possible to add to the docker image as well, so that it starts both server and client?

dandersch:
@Migs3 Not via this user service - the docker image doesn't (and shouldn't) use systemd. My use case was more for desktop machines that want to use only the client so that notifications can pop up with `notify-send` (without configuring anything in `/etc` like the `override.conf` or `client.yml`). If you want this functionality with a docker container, I think you would have to do what the `override.conf` did, i.e. include something like
```docker-compose.yaml
    environment:
      - DISPLAY=:0
      - DBUS_SESSION_BUS_ADDRESS=unix:path=/run/user/1000/bus
    volumes:
      - /run/user/1000/bus:/run/user/1000/bus
    command:
    #  - serve
    - subscribe
```

binwiederhier:
I would love to take it, but when I run your package install, I get this:

```
sudo dpkg -i dist/ntfy_v2.11.0-next_linux_amd64.deb 
Selecting previously unselected package ntfy.
(Reading database ... 267047 files and directories currently installed.)
Preparing to unpack .../ntfy_v2.11.0-next_linux_amd64.deb ...
Unpacking ntfy (2.11.0~next) ...
Setting up ntfy (2.11.0~next) ...
Failed to connect to bus: No medium found <<<<<<<<<<<<<<<
Failed to connect to bus: No medium found
```

binwiederhier:
And yes: I realize that I am over a year too late.

dandersch:
Thanks for taking a look.

I'm on Arch, so I don't know how to test `deb` packages, but I noticed it doesn't make sense to have `systemctl --user` commands in the postinstall script if it gets invoked with `sudo`.

I replaced the calls with a message to the user after installation:
```
------------------------------------------------------------------------
ntfy includes a systemd user service.
To enable it, run following commands as your regular user (not as root):

  systemctl --user enable ntfy-client.service
  systemctl --user start ntfy-client.service
------------------------------------------------------------------------
```

PR #1014, opened 2024-01-30
Docs: Enhance Watchtower example

lennart-m:
Document usage of environment variable `WATCHTOWER_NOTIFICATION_SKIP_TITLE` for the Watchtower container so the provided title will not be overwritten.

I tested the example on my system. Without this parameter the title will be ignored, and Watchtower will set its own title.


binwiederhier:
Thank you

PR #1016, opened 2024-02-05
Remove dependence on mkdocs-simple-hooks

Tom-Hubrecht:
Since mkdocs v1.4, the hooks are a native feature


binwiederhier:
Very cool thanks!

PR #1031, opened 2024-02-22
fix(make): Build static binaries, to fix runtime errors in containers.

rjshrjndrn:


binwiederhier:
goreleaser sets CGO_ENABLED; what are you trying to solve?

(I'm closing this, because there are no issues; let me know what the problems are and we can discuss further)

PR #1032, opened 2024-02-22
Clarify that Content-Length is required in the HTTP header

CrossVR:
Without it your message will be empty and the message will be replaced by "triggered".

I initially missed this because normally this happens when using a GET request instead of POST, so it took a bit of time to figure out I had forgotten the Content-Length header.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -40,6 +40,7 @@ simple message using a POST request:
     ``` http
     POST /mytopic HTTP/1.1
     Host: ntfy.sh
+    Content-Length: 22
     
     Backup successful 😀
     ```


binwiederhier:
You don't really need the Content-Length header at all. Here's an example with `nc`:

```
$ echo -e "POST /mytopic1 HTTP/1.1\r\nHost: ntfy.sh\r\nConnection: close\r\n\r\nTest with netcat" | nc -v ntfy.sh 80
Connection to ntfy.sh (159.203.148.75) 80 port [tcp/http] succeeded!
HTTP/1.1 200 OK
Server: nginx
Date: Thu, 07 Mar 2024 15:15:07 GMT
Content-Type: application/json
Content-Length: 120
Connection: close
Access-Control-Allow-Origin: *

{"id":"jM0rj01rYhB0","time":1709824507,"expires":1709867707,"event":"message","topic":"mytopic1","message":"triggered"}
```

Please note that it also works with out the `Connection: close` though then it'll expect more requests.

CrossVR:
Your own example shows that you do need the Content-Length header, look at the response you got from the server:

```
{
    "id":"jM0rj01rYhB0",
    "time":1709824507,
    "expires":1709867707,
    "event":"message",
    "topic":"mytopic1",
    "message":"triggered"
}
```
The message you sent was "Test with netcat", yet it was replaced by "triggered" just as expected.

binwiederhier:
Whoa 😮

binwiederhier:
My apologies. That's crazy. I guess I've never sent a raw HTTP request... 

I think we'll have to update all examples though not just this one.

CrossVR:
No worries, this is also often missed because some utilities will implicitly add a Content-Length header if it's missing. But if you send a truly raw HTTP request then it is necessary to manually add it, so it's useful to add to the docs to avoid confusion.

binwiederhier:
While technically correct, I think this is merely cosmetic. So I reserve the right to close this. Thank you for reporting.

PR #1044, opened 2024-03-06
Update Danish translations

tripleee:
Avoid compound errors

This is based on my work in Swedish, but I am not fluent in Danish, so you should probably get this reviewed by a native speaker (though ideally not someone who habitually makes compound errors themself).

"Reserverede emner tabel" and "Fjern telefon opkald" sound like compound errors, too, but I am not confident enough to propose fixes for them.


binwiederhier:
Thank you

PR #1062, opened 2024-03-21
Update publish.md (PowerShell 7+ Access Token Example)

theparadox1083:
Corrected PowerShell 7+ Parameter from Authorization to Authentication 
Converted Token text to SecureString Inline - token must be of type SecureString 
Added comment noting the Token parameter must be a SecureString


binwiederhier:
Awesome thank you!

PR #1082, opened 2024-04-04
Fix username regex

wunter8:
Now you can include a "+" in your username (for example, in an email alias).

I also added some code to convert the user/types.go ErrInvalidArgument into an errHTTPBadRequestInvalidArgument so that the error message in the web UI says "invalid request: invalid argument" instead of "internal server error."

I feel like that part of the PR isn't super clean, so you might want to rewrite it (all 6 lines)

### DIFF
--- a/server/errors.go
+++ b/server/errors.go
@@ -122,7 +122,7 @@ var (
 	errHTTPBadRequestTemplateInvalid                 = &errHTTP{40043, http.StatusBadRequest, "invalid request: could not parse template", "https://ntfy.sh/docs/publish/#message-templating", nil}
 	errHTTPBadRequestTemplateDisallowedFunctionCalls = &errHTTP{40044, http.StatusBadRequest, "invalid request: template contains disallowed function calls, e.g. template, call, or define", "https://ntfy.sh/docs/publish/#message-templating", nil}
 	errHTTPBadRequestTemplateExecuteFailed           = &errHTTP{40045, http.StatusBadRequest, "invalid request: template execution failed", "https://ntfy.sh/docs/publish/#message-templating", nil}
-	errHTTPBadRequestInvalidArgument                 = &errHTTP{40046, http.StatusBadRequest, "invalid request: invalid argument", "", nil}
+	errHTTPBadRequestInvalidUsername                 = &errHTTP{40046, http.StatusBadRequest, "invalid request: invalid username", "", nil}
 	errHTTPNotFound                                  = &errHTTP{40401, http.StatusNotFound, "page not found", "", nil}
 	errHTTPUnauthorized                              = &errHTTP{40101, http.StatusUnauthorized, "unauthorized", "https://ntfy.sh/docs/publish/#authentication", nil}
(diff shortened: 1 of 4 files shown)


server/server_account.go
```diff
@@ -37,7 +37,11 @@ func (s *Server) handleAccountCreate(w http.ResponseWriter, r *http.Request, v *
 	}
 	logvr(v, r).Tag(tagAccount).Field("user_name", newAccount.Username).Info("Creating user %s", newAccount.Username)
 	if err := s.userManager.AddUser(newAccount.Username, newAccount.Password, user.RoleUser); err != nil {
-		return err
+		if err.Error() == "invalid argument" {
```
binwiederhier, line 40:
```suggestion
		if errors.Is(err, ErrInvalidArgument) {
		  return errHTTPBadRequestInvalidUsername
		}
		return err	
```

server/errors.go
```diff
@@ -122,6 +122,7 @@ var (
 	errHTTPBadRequestTemplateInvalid                 = &errHTTP{40043, http.StatusBadRequest, "invalid request: could not parse template", "https://ntfy.sh/docs/publish/#message-templating", nil}
 	errHTTPBadRequestTemplateDisallowedFunctionCalls = &errHTTP{40044, http.StatusBadRequest, "invalid request: template contains disallowed function calls, e.g. template, call, or define", "https://ntfy.sh/docs/publish/#message-templating", nil}
 	errHTTPBadRequestTemplateExecuteFailed           = &errHTTP{40045, http.StatusBadRequest, "invalid request: template execution failed", "https://ntfy.sh/docs/publish/#message-templating", nil}
+	errHTTPBadRequestInvalidArgument                 = &errHTTP{40046, http.StatusBadRequest, "invalid request: invalid argument", "", nil}
```
binwiederhier, line 125: errHTTPBadRequestInvalidUsername?

user/types.go
```diff
@@ -241,7 +241,7 @@ const (
 )
 
 var (
-	allowedUsernameRegex     = regexp.MustCompile(`^[-_.@a-zA-Z0-9]+$`)     // Does not include Everyone (*)
+	allowedUsernameRegex     = regexp.MustCompile(`^[-_.+@a-zA-Z0-9]+$`)    // Does not include Everyone (*)
```
binwiederhier, line 244: Can you check if there is a regex in the frontend too somewhere?

binwiederhier (approved):

wunter8:
I'm planning to get to this today

server/errors.go
wunter8: Right now, 'ErrInvalidArgument' is returned for invalid usernames, topics, roles, and tiers.

But, in this section of the code specifically, we're only checking for usernames, so the 'ErrInvalidArgument' will always mean 'errHTTPBadRequestInvalidUsername'

user/types.go
wunter8: I looked and didn't find anything

PR #1083, opened 2024-04-07
chore: remove repetitive words

alwayshang:


binwiederhier:
Your changes are correct, though you joined GitHub this month and your profile is private. That implies to me that you are a karma farming account. I hope this is not correct, but for now I won't accept this PR.

PR #1086, opened 2024-04-14
docs(integrations): add proxmox-ntfy

ibacalu:
Python script that monitors Proxmox tasks and sends notifications using Ntfy.
More details [here](https://github.com/qtsone/proxmox-ntfy)


binwiederhier:
Thanks

Issue #303, opened 2022-06-01
Update/delete notifications


blacklight:
A [similar issue](https://github.com/binwiederhier/ntfy/issues/43) was open a while ago, but it didn't get traction for some reason.

The ability to update notifications is so far the only feature that forces me to still use AutoNotification and the Tasker environment instead of ntfy.

I have a multi-host and multi-room music setup, and whenever a new track is playing I receive an actionable notification that allows me to easily control the playback.

AutoNotification allows editing a notification given its ID (even a total overwrite/replace is fine, so we don't have to mess with the complexity of partial updates). And that's exactly the feature I need: if every music device created a new notification for each track that is played the notification bar will quickly get crammed, especially if there is no programmatic way to dismiss stale notifications. Notifications that report a progress status are another quite interesting use case that isn't possible right now.

binwiederhier:
There's also #44 for progress. That is trivial to implement once messages can be updated.

BlackLight:
> I want the API to be simple and ideally the request to update a message is the same that it is to initially publish it.

Something like `POST`/`PUT` without an `id` to create a notification and `PUT` with an `id` to update an existing one? `id` could then be passed either on the query string or the JSON payload.

KawaiiZapic:
It's great to have this feature.   
we can set an id(or type) for a message, and replace the old message if id is same.

mrherman:
For the API, we could consider something similar to Home Assistants companion app's notification replace/clear system.  https://companion.home-assistant.io/docs/notifications/notifications-basic/#replacing  For that, the user/client provides a tag/id name for the notifications.  On the receiver side, if no notification with that tag exists then it creates it, otherwise it deletes/replaces it.  It is up to the user to make the tag super unique or generic depending on their use case.

Just to add, I would really like this feature too.  I pop up an image of the outdoor cam when doorbell is rang, but it is only relevant until no motion is detected (for me to decide to answer it or not), after that I would like to clear the notification.  Also, sticky notification (if we get them) that has some data that is periodically updated, I can update the notification when the data is changed.

julianlam:
Another +1 for this suggestion, if I may :smile: 

We would use this in NodeBB for replacing notifications with updated text or to remove stale notifications. For example, a notification that a post needs review is no longer relevant if another admin has approved/denied it, and so the notification automatically gets rescinded in NodeBB.

[...]

PR #1142, opened 2024-06-30
[feature] implement scheduled message unpublishing via `/topic/{messageId}`

GamerGirlandCo:
Test cases are included.
Related: #303

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -77,6 +77,7 @@ var (
 	wsPathRegex            = regexp.MustCompile(`^/[-_A-Za-z0-9]{1,64}(,[-_A-Za-z0-9]{1,64})*/ws$`)
 	authPathRegex          = regexp.MustCompile(`^/[-_A-Za-z0-9]{1,64}(,[-_A-Za-z0-9]{1,64})*/auth$`)
 	publishPathRegex       = regexp.MustCompile(`^/[-_A-Za-z0-9]{1,64}/(publish|send|trigger)$`)
+	messagePathRegex       = regexp.MustCompile(`^/[-A-Za-z-0-9]{1,64}/([A-Za-z0-9]{12})$`)
 
 	webConfigPath                                        = "/config.js"
 	webManifestPath                                      = "/manifest.webmanifest"
@@ -537,6 +538,8 @@ func (s *Server) handleInternal(w http.ResponseWriter, r *http.Request, v *visit
 		return s.limitRequests(s.authorizeTopicRead(s.handleTopicAuth))(w, r, v)
 	} else if r.Method == http.MethodGet && (topicPathRegex.MatchString(r.URL.Path) || externalTopicPathRegex.MatchString(r.URL.Path)) {
 		return s.ensureWebEnabled(s.handleTopic)(w, r, v)
+	} else if r.Method == http.MethodDelete && (messagePathRegex.MatchString(r.URL.Path)) {
+		return s.limitRequestsWithTopic(s.authorizeTopicWrite(s.handleMessageUnpublish))(w, r, v)
 	}
 	return errHTTPNotFound
 }
@@ -556,6 +559,25 @@ func (s *Server) handleTopic(w http.ResponseWriter, r *http.Request, v *visitor)
 	}
 	r.URL.Path = webAppIndex
 	return s.handleStatic(w, r, v)
+}
+
+func (s *Server) handleMessageUnpublish(w http.ResponseWriter, r *http.Request, v *visitor) error {
+	segments := strings.Split(r.URL.Path, "/")
+	msg, err := s.messageCache.Message(segments[len(segments)-1])
+	if err != nil {
+		if err == errMessageNotFound {
+			return errHTTPNotFound
+		}
+		return err
+	}
+	if time.Now().Unix() >= msg.Time {
+		httpErr := errHTTPBadRequest
+		alreadyPublished := httpErr.Wrap("Message \"%s\" has already been published", msg.ID)
+		s.handleError(w, r, v, alreadyPublished)
+		return alreadyPublished
+	}
+	s.messageCache.DeleteMessages(msg.ID)
+	return s.writeJSON(w, msg)
 }
 
 func (s *Server) handleEmpty(_ http.ResponseWriter, _ *http.Request, _ *visitor) error {
--- a/server/server_test.go
+++ b/server/server_test.go
@@ -7,8 +7,6 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
-	"golang.org/x/crypto/bcrypt"
-	"heckel.io/ntfy/v2/user"
 	"io"
 	"net/http"
 	"net/http/httptest"
@@ -21,6 +19,9 @@ import (
 	"sync/atomic"
 	"testing"
 	"time"
+
+	"golang.org/x/crypto/bcrypt"
+	"heckel.io/ntfy/v2/user"
 
 	"github.com/SherClockHolmes/webpush-go"
 	"github.com/stretchr/testify/require"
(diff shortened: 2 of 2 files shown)


binwiederhier:
This is awesome. I will 100% get this in for the next release. 

I already rebased this on top of the main branch locally, but I have a question. 

Given the latest feature (https://docs.ntfy.sh/publish/#updating-deleting-notifications) already implements the DELETE endpoint, my thinking was this:

- If the message is not yet published, delete it from the cache (your code)
- If it is already published, emit a message_delete event (logic in main)

Analogously, publishing to /topic/message-id should update the message in the cache if it's not published yet (assuming it has a Delay), or otherwise treat it as a new message?! I'm a bit fuzzy on what should happen here. 

@wunter8 Thoughts?

binwiederhier:
I'm very excited about implementing updates on unpublished messages, because that can serve as a dead man's switch

wunter8:
I think what you described, binwiederhier, is the same as what I originally considered here: https://github.com/binwiederhier/ntfy/issues/303#issuecomment-2927423013

Copying some thoughts I had previously:

I think we should only keep 1 scheduled message for a particular sequence_id at a time. So a first message can be scheduled with `sequence_id: reminder`. Then, if we want to update that (push the reminder into the future, change the reminder message, etc.), we can POST a new message with `sequence_id: reminder` and whatever fields we want (message, delay, attachments, etc.). The previous revision would be removed from the DB (I'm thinking deleted entirely), and the new revision would be added/scheduled to be sent in the future.

I think deleting entirely might be better than modifying it's timestamp field or published field because if we change the DB to show the message was published and don't send anything to the client (since it wasn't actually sent), if the client polls the server for all messages in the cache for a topic, they'd get a bunch of scheduled revisions that were never supposed to be sent.

If we delete the revisions, using the subscriber API, you could still check for scheduled messages on a particular topic, but you wouldn't be able to see a history of revisions for the scheduled message, only the most recent.

However, I'm now considering the scenario when you send (assuming all these have the same `sequence_id`) message `A` then immediately send message `B` with a 10 minute delay. Then there are two messages in the cache for that `sequence_id`: `A` which has been published and `B` which hasn't been published. If you then send `DELETE /topic/sequence_id`, what should happen?

Option A: the scheduled message is deleted/unscheduled
Option B: the published message A is deleted on client devices
Option C: both A and B
Option D: we add a parameter to the DELETE request that lets the user choose between options A, B, and C??

I think clearing a message doesn't have the same problem because you can't clear something that hasn't sent yet. So I think just DELETE has these extra considerations

wunter8:
One way to distinguish between deleting an existing message and deleting a scheduled message could be to require setting a delay when deleting a scheduled message (even though the delete will happen immediately)

So

```
DELETE /topic/sequence_id
```

would delete an existing/already sent message with `sequence_id`

And

```
DELETE /topic/sequence_id
Delay: 10 min
```

would delete a scheduled/delayed message with `sequence_id`

binwiederhier:
Your comment is great. It's very complicated. My brain is not working right because I didn't sleep. I will re-read later/tmr and respond.

I think we somehow need to restrict sequences with scheduled messages, so that there can never be more than one scheduled message in a sequence in the DB. 

This is what I was trying to come up with:

```
DELETE /mytopic/myseq
  m := s.cache.MessageBySID(myseq)
  if m.Published {
    return {sequence_id:myseq,event:message_delete,...}
  } else {
    s.cache.DeleteBySID(myseq) // Does this hard-delete all messages in the seq?
    return m
  }
 
POST /mytopic/myseq -- update message 
  m := s.cache.MessageBySID(myseq)
  if !m.Published {
    s.cache.DeleteBySID(myseq) // Does this hard-delete all messages in the seq?
  }
  return {id:abc...,sequence_id:myseq,event:message,...}
```

binwiederhier:
@wunter8  I tried re-reading your comment and even with "awake-brain" I still have trouble understanding it.

So let me try to analyze it one paragraph at a time:

> I think we should only keep 1 scheduled message for a particular sequence_id at a time. So a first message can be scheduled with `sequence_id: reminder`. Then, if we want to update that (push the reminder into the future, change the reminder message, etc.), we can POST a new message with `sequence_id: reminder` and whatever fields we want (message, delay, attachments, etc.). The previous revision would be removed from the DB (I'm thinking deleted entirely), and the new revision would be added/scheduled to be sent in the future.

Yes! So basically, if you try to update or delete a **scheduled message** with a sequence ID, we must look it up and delete the previous versions of it:

```
DELETE /mytopic/myseq
  m := s.cache.MessageBySID(myseq)
  if !m.Published {
    s.cache.DeleteBySID(myseq) // Does this hard-delete all messages in the seq?
  }

POST /mytopic/myseq -- update message 
  m := s.cache.MessageBySID(myseq)
  if !m.Published {
    s.cache.DeleteBySID(myseq) // Does this hard-delete all messages in the seq?
    return {id:newIDabc...,sequence_id:myseq,event:message,...}
  }
```


> I think deleting entirely might be better than modifying it's timestamp field or published field because if we change the DB to show the message was published and don't send anything to the client (since it wasn't actually sent), if the client polls the server for all messages in the cache for a topic, they'd get a bunch of scheduled revisions that were never supposed to be sent.

Agreed. We must ensure that only one **scheduled message per sequence ID** exists in the database.

> If we delete the revisions, using the subscriber API, you could still check for scheduled messages on a particular topic, but you wouldn't be able to see a history of revisions for the scheduled message, only the most recent.

So if we do `/mytopic/json?poll=1&scheduled=1`, the previously scheduled message would disappear. That's what you meant, right? If so, yes, that is correct, because we'll delete the previously scheduled message from the DB.

> However, I'm now considering the scenario when you send (assuming all these have the same `sequence_id`) message `A` then immediately send message `B` with a 10 minute delay. Then there are two messages in the cache for that `sequence_id`: `A` which has been published and `B` which hasn't been published. If you then send `DELETE /topic/sequence_id`, what should happen?
> 
> Option 1: the scheduled message is deleted/unscheduled
> Option 2: the published message A is deleted on client devices
> Option 3: both 1 and 2
> Option 4: we add a parameter to the DELETE request that lets the user choose between options 1, 2, and 3??

(I renamed the options to 1..4 to avoid confusion with message A + B.)

This is the scenario you described:

```
curl -d created ntfy.sh/mytopic/myseq1  ---> (a)
curl -d updated -H "Delay: 30s" ntfy.sh/mytopic/myseq1 --> (b)
curl -X DELETE ntfy.sh/mytopic/myseq1  ---> What to do??
```

Honestly, I don't know. This is clearly ambigious. Maybe unpublish is a separate API endpoint, like `POST /mytopic/seq1/cancel`? Or we also make the `message_delete` event a different endpoint??!

> I think clearing a message doesn't have the same problem because you can't clear something that hasn't sent yet. So I think just DELETE has these extra considerations

:+1:

wunter8:
Yes, you understood it all correctly!

Did you see my second message with a possible way of resolving the ambiguity?

binwiederhier:
> Did you see my second message with a possible way of resolving the ambiguity?

I did! And I hate it, haha.

I just had a realization regarding the 4 options. `DELETE ntfy.sh/mytopic/myseq1` deletes a sequence, not a single message, so we should do both: hard delete all scheduled messages in the db AND send out a message_delete event. Everything else would lead to confusion I think and would be perceived as a bug.

How about this:
```
DELETE /mytopic/myseq
  s.cache.DeleteScheduledBySID(myseq)
  m := {sequence_id:myseq,event:message_delete,...}
  m.cache.AddMessage(m)
  return m
 
POST /mytopic/myseq -- update message 
  s.cache.DeleteScheduledBySID(myseq)
  m := {id:abc...,sequence_id:myseq,event:message,...}
  m.cache.AddMessage(m)
  return m
```

wunter8:
Deleting a sent message feels like it should be different from deleting a scheduled message. But I can also understand your perspective since we're dealing with whole "sequence_ids" and not individual messages

binwiederhier:
This works wonderfully. 100% written by Cursor: https://github.com/binwiederhier/ntfy/pull/1556

> I'd like to implement updating and deleting of unpublished (scheduled) messages. Conceptually, I'd like to fit this into the concept of a "sequence id". 
> 
> - When updating an (scheduled) message, I'd like to delete the existing scheduled message with the same sequence id from the database, and add a new message with the same sequence id. This logic goes into the existing handlePublish() endpoint handler.
> 
> - When deleting a (scheduled) message, i'd like to delete the existing scheduled message with the same sequence id (like above) from the database, and emit a message_delete event (like it happens already today). This should go in the handleDelete handler.
> 
> Pseudo code:
> 
> DELETE /mytopic/myseq
>   s.cache.DeleteScheduledBySID(myseq)
>   m := {sequence_id:myseq,event:message_delete,...}
>   m.cache.AddMessage(m)
>   return m
>  
> POST /mytopic/myseq -- update message 
>   s.cache.DeleteScheduledBySID(myseq)
>   m := {id:abc...,sequence_id:myseq,event:message,...}
>   m.cache.AddMessage(m)
>   return m

I read the code and it all looks sound. I need to update the attachment deletion for unpublished messages, then I'll merge it and deploy it to staging.

binwiederhier:
Superseded by https://github.com/binwiederhier/ntfy/pull/1556

Thanks @GamerGirlandCo for implementing this. Due to the complexity of this (with the other "update and delete notifications" code since added), I can't merge this. I gave you credit in the release notes!

PR #1164, opened 2024-08-08
Support SMTP Auth Plain for event publishing

bishtawi:
Hi, very cool project. I am currently looking into hosting the server and I need to authenticate the services that are publishing ~~notifications~~ emails. I read the docs and saw that you can include an access token as part of the email (`ntfy-$topic+$token@ntfy.sh`) but unfortunately that solution does not work for my usecase. Access tokens are sensitive information and with my set up, email addresses are not. Including access tokens as part of the email address is not a feasible solution.

SMTP has authentication protocols and a common one is AUTH PLAIN which your SMTP dependency (go-smtp) supports. We can hook up the SMTP AUTH PLAIN command to forward the credentials to the ntfy server and authenticate using basic auth.

I was honestly surprised at how easy it was to implement this, but I believe these are all the changes necessary. Let me know if there is anything else required.

### DIFF
--- a/server/smtp_server.go
+++ b/server/smtp_server.go
@@ -70,15 +70,20 @@ func (b *smtpBackend) Counts() (total int64, success int64, failure int64) {
 
 // smtpSession is returned after EHLO.
 type smtpSession struct {
-	backend *smtpBackend
-	conn    *smtp.Conn
-	topic   string
-	token   string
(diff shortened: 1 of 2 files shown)


binwiederhier:
This looks straight forward. Thank you. I'll add some docs around it.

PR #1198, opened 2024-10-11
Use /usr/local/bin instead of /usr/bin

allan2:
This commit makes the placement of binaries consistent on Linux. ntfy currently recommends /usr/local/bin for amd64 Linux but not for arm Linux.

- replaced all instances of `/usr/bin/ntfy` with `/usr/local/bin/ntfy`
- replaced two instances of `/usr/bin` with `/usr/local/bin`
    - in Dockerfile and Dockerfile-arm, the line is changed to `COPY ntfy /usr/local/bin`
    
This PR does not add anything to move binaries or remove them from the old path.


binwiederhier:
I appreciate the PR. Thank you for the effort, but I don't think I want to do that. It'll break too many things downstream.

PR #1213, opened 2024-11-03
add: owncast-ntfy

HolgerHuo:
owncast-ntfy proxies [Owncast's Webhook Events](https://owncast.online/thirdparty/webhooks/) to ntfy messages


wunter8:
Nice! Thanks for sharing!

I looked briefly at the code, and it seems you're just converting from Owncast's JSON format to ntfy's JSON format, right? If so, there's a new message templating feature that can do this natively in ntfy: https://docs.ntfy.sh/publish/#message-templating

Right now, it's limited to notification titles and messages. So if your app converts other fields (e.g., tags, priority, etc.) (I didn't look close enough at your code 😅), that part can't be done in ntfy directly yet

HolgerHuo:
Hi! Thank you for sharing this templating feature which i didn't think of its usage before. This is excellent for integrating many 3rd party applications. 

I did utilize `tags` feature, but they were all static ones, so a simple query string should do the trick. 

Maybe we could add a new section in the docs for sharing message templates so that more people could utilize this feature?

wunter8:
That might be good. Maybe we can add a subsection under Examples and link to it from here: https://docs.ntfy.sh/publish/#message-templating

PR #1218, opened 2024-11-11
Add support for server templates

UiP9AV6Y:
instead of sending title/message templates as part of the publishing request, systems can simply reference existing templates from the server. this has two advantages
a) publish URLs become shorter
b) publish URLs become easier to reuse

the changes backwards-compatible, as the `tpl` parameter continues to support truthy boolean values.
the feature has to be enabled on the server. available templates and their content are exposed via new API endpoints.


wunter8:
Wow! Thanks for working on this! Having message templates on the server is one of the features we talked about in the past, but obviously we haven't implemented it yet.

Binwiederhier will be the one to decide whether to merge this, but here are some of my thoughts:
1. It's great that this is backwards-compatible
2. I don't love that the feature is all or nothing: title and message both online or both server templates. I'm thinking mostly for titles, which are typically short and might not even depend on the contents of the JSON body. For example, having to create a file on the server just to have a title with "New Grafana Alert" (or any other fixed string) would be unfortunate.

I don't know the best way to remedy this, though. I considered something like `tpl=server,inline` for a server message and inline title. Or `tpl=1&message=server:grafana&title=New+Grafana+Alert`. Looking toward the future, though, we have discussed making it possible to template all message fields (such as priority and tags) based on the JSON body. I don't know what that looks like and how it combines with this. 

3. I don't use Golang very much. I saw some different styles between your code and the existing ntfy code. Not sure which is better 🤷‍♂️
4. You seemed to do a good job adding additional tests while keeping the existing template tests 👍
5. Thanks for keeping security concerns in mind like path traversals!

binwiederhier:
@wunter8 I don't hate having server side templates. Not saying you could work on this, I'd be curious to have this :-D

wunter8:
How do you envision that working?

Are you fine with the "all or nothing" approach? Do you want to support a combination of query param templates and server file templates in a single request (e.g., the title template is a query param and the message template is a file on the server)?

binwiederhier:
I don't hate this: `https://ntfy.sh/mytopic?tpl=server&title=hello_world&message=grafana_body`

Though it'd be nicer to have it work less detached from the other template feature, maybe `https://ntfy.sh/mytopic?tpl=grafana`

If the template could define/predefine all the ntfy fields that'd be good.

Maybe the template is a yml file:

``` yaml
ntfy-template:
  priority: {{if bla}}3{{else}}low{{endif}}
  title: Grafana alert {{.Title}}
  message: |
    {{if .Priority} ...
```

No clue if/how that would work, but this is nicer than the weird "use message/title as template body"

Issue #905, opened 2023-10-01
Prefer Cloudflare IP header instead of X-Forwarded-For


ChokunPlayZ:
:bulb: **Idea**
I think there's not many configs like mine where the setup involves multiple reverse proxy (in my case NGINX and Cloudflare)
the current "behind proxy" setting cause issue because
NGINX attach 2 IP in the forwarded for header, ntfy always prefer the wrong one (the last one)
instead there's the Cloudflare's `Cf-Connecting-Ip` which is more accurate than the `X-Forwarded-For` header
it would be great if this could be added

or an alternative fix, if there is multiple IP in the `X-Forwarded-For` header always prefer the public IP over private ones
because in my case here's how the header looks like: `X-Forwarded-For: <mypublicip>, 172.17.0.1(docker)`

:computer: **Target components**
ntfy server

binwiederhier:
I understand that the X-Forwarded-For handling is rudimentary. ntfy picks the rightmost address, since that is the most secure way to pick a trustworthy address. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For for details.

To implement a proper X-Forwarded-For handling, we'd need to be able to configure a list of trustworthy proxies, e.g. `proxies: 172.17.0.1`. These proxies would be ignored in the selection of the IP address.

Adding a vendor-specific header like `Cf-Connecting-Ip` doesn't seem like a great idea, IMHO.

hipur:
Maybe X-Real-IP could be beneficial, since it's easier to overwrite without reverse proxies adding another IP to the right. e.g.: traefik

ojio-san:
Hello there,

I'm sorry to revive such an old topic, but this issue is open, so I would like to add something to this, as I have multiple proxies there, so I have the same issue as @ChokunPlayZ 

@binwiederhier 
> I understand that the X-Forwarded-For handling is rudimentary. ntfy picks the rightmost address, since that is the most secure way to pick a trustworthy address. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For for details.

I would like to know where did you find that the rightmost address is the most secure way and the trustworthy address ?

If I'm reading right here : https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For#client

`This means that the rightmost IP address is the IP address of the most recent proxy and the leftmost IP address is the address of the originating client`

[...]

PR #1252, opened 2025-01-06
Adding proxy client IP header override functionality

pixitha:
Based on this #905, I added the ability to specify the header for the proxy server to send the real client IP, if you don't want to use XFF header.

Was able to test and build and deploy this to fly.io and it now correctly identifies the client ip.

Still need to refactor it a bit more I think, maybe add the option to specify the subnet for the proxy server for still using XFF with a bit more logic?


binwiederhier:
I'd be happy to accept this PR, though it is (as you say) a tad messy. Did you want to clean this up some more?

PR #1257, opened 2025-01-10
Add 'Ntfyhook' to integrations

rocord01:
[Ntfyhook](https://git.rocord.dev/rocord01/ntfyhook) is a proxy/compatibility layer between Discord webhooks and Ntfy


binwiederhier:
This says bad gateway

PR #1267, opened 2025-01-26
Api change user password

wunter8:
I'm not sure if this is the best way to implement this, but I added a `force` param to the "add user" request.

If `force == true`, it will override the password of an existing user (instead of it causing an error saying that a user already exists).

This will let you change a user's password and preserve the user ID in the database (instead of deleting a user and then creating a new one with the same username)

The "add user" endpoint already requires admin credentials to use, and admins can create and delete users, so I didn't think changing a user's password would be much different.

Admins CANNOT change the password of another admin.

### DIFF
--- a/server/server_admin.go
+++ b/server/server_admin.go
@@ -39,7 +39,7 @@ func (s *Server) handleUsersGet(w http.ResponseWriter, r *http.Request, v *visit
 }
 
 func (s *Server) handleUsersAdd(w http.ResponseWriter, r *http.Request, v *visitor) error {
-	req, err := readJSONWithLimit[apiUserAddRequest](r.Body, jsonBodyBytesLimit, false)
+	req, err := readJSONWithLimit[apiUserAddOrUpdateRequest](r.Body, jsonBodyBytesLimit, false)
 	if err != nil {
 		return err
 	} else if !user.AllowedUsername(req.Username) || req.Password == "" {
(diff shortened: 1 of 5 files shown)


binwiederhier:
This is just an UPSERT, right? Traditionally PUT is used to update and POST to create, according to chatgpt, so I fucked up the original API.. I am not a fan of the force thing. But if you really need it I'll merge it

wunter8:
Yes, traditionally PUT is to update and POST to create. I think I may have noticed a discrepancy when I worked on this, but I'm not at my desk right now to remind myself.

I don't need this. It came up as I was supporting someone on Discord. You can change a password via the command line, so I thought it'd be okay to add it to the API. The alternative is to delete the old user with the existing API and create a new user with the existing API. However, that would result in a new user ID, which I thought might cause problems. Right now, I can't think of any problems specifically, though, since messages are sent to topics and not users.

I'm okay holding off on this for now. We can discuss a different request pattern for changing passwords if you want to make that an option at all.

binwiederhier:
We could just make the PUT always upsert, and make a POST that just creates the user. It's a slight change in the API but it should be fine

wunter8:
Sounds good to me. Will we require a POST before a PUT? Or can you PUT directly?

binwiederhier:
Nah, we could leave the PUT as an UPSERT and just make the POST behave like the current PUT.

So:
- POST: Creates user, fails if user already exists
- PUT: Creates user, updates user if it already exists

wunter8:
Want me to do that?

binwiederhier:
Sure :-)

server/server_admin.go
```diff
@@ -70,6 +70,44 @@ func (s *Server) handleUsersAdd(w http.ResponseWriter, r *http.Request, v *visit
 	}
 	return s.writeJSON(w, newSuccessResponse())
 }
+func (s *Server) handleUsersUpdate(w http.ResponseWriter, r *http.Request, v *visitor) error {
+	req, err := readJSONWithLimit[apiUserAddOrUpdateRequest](r.Body, jsonBodyBytesLimit, false)
+	if err != nil {
+		return err
+	} else if !user.AllowedUsername(req.Username) || req.Password == "" {
```
binwiederhier, line 77: This implies that this API is only for changing the password. What if we only want to change the tier?

 `{"username": "ben", "tier": "pro"}`

I know that the API isn't originally for that, and it's undocumented, but why not :-)

binwiederhier (approved):

wunter8:
You can now modify the password and/or tier using PUT

PR #1287, opened 2025-03-05
Handle anonymous read restrictions by sending a poll_request event

barart:
If a topic does not allow anonymous reads, this change ensures that we send a "poll_request" event instead of relaying the message via Firebase. Additionally, we include generic text in the title and body/message. This way, if the client cannot retrieve the actual message, the user will still receive a notification, prompting them to update the client manually.

### DIFF
--- a/server/server_firebase.go
+++ b/server/server_firebase.go
@@ -175,13 +175,26 @@ func toFirebaseMessage(m *message, auther user.Auther) (*messaging.Message, erro
 		} else {
 			// If anonymous read for a topic is not allowed, we cannot send the message along
 			// via Firebase. Instead, we send a "poll_request" message, asking the client to poll.
+			//App function needs all the data to create a message object, if not, it fails,
+			//so we set it but put a placeholders to not to send the actual message
+			//but generic title and message instead, we also add the poll_id so client knowns
+			//what message is goint to "decode" (retrieve)
 			data = map[string]string{
-				"id":    m.ID,
-				"time":  fmt.Sprintf("%d", m.Time),
-				"event": pollRequestEvent,
-				"topic": m.Topic,
-			}
-			// TODO Handle APNS?
+                                "id":    m.ID,
+                                "time":  fmt.Sprintf("%d", m.Time),
+                                "event": pollRequestEvent,
+                                "topic": m.Topic,
+                                "priority":     fmt.Sprintf("%d", m.Priority),                                            
+                                "tags":         strings.Join(m.Tags, ","),                                                
+                                "click":        m.Click,                                                                  
+                                "icon":         m.Icon,                                                                      
+                                "title":        "Private",                                                                     
+                                "message":      "Message",                                                                
+                                "content_type": m.ContentType,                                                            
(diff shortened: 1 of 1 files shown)


binwiederhier:
Holy shit, is this the issue why some users do not receive message on iOS if they have a Pro subscription?

wunter8:
@binwiederhier Yes, I'm pretty sure this is the source of/fix for that bug! I think I worked through this with barart on Discord

barart:
@binwiederhier I’m not sure what features Pro subscription users have, but this fix addresses the issue with the iOS app and APNS when the topic is private. Additionally, this other fix is also required in the iOS app:
https://github.com/binwiederhier/ntfy-ios/pull/20

I host my own ntfy server, compiled my own version of the ntfy iOS app, and use my own APNS configuration via my Apple Developer account, so I’m not entirely sure what the Pro subscription includes or how this fix might impact it. But I’d be happy to help investigate if you can guide me on what issue your Pro users are facing.

And yep, we worked on this with @wunter8 on Discord that day, he gave me the clue that led to the fix.

wunter8:
That's exactly the issue binwiederhier is talking about: Pro subscriptions let people make a certain number of topics private on the public ntfy.sh instance. However, once they make the topic private, they don't receive push notifications through the iOS app anymore

barart:
So yes, this will fix the problem 👌🏻

binwiederhier:
So I don't think I agree with the tags and such being part of the poll request. I'll merge your PR, but then later empty-out these fields. I don't want to leak data.

binwiederhier:
Thank you. I hope this works!

binwiederhier:
My brain is not fully functioning anymore, so I'll have to re-read what you wrote tomorrow, but please look at what I did here and tell me if you think this would still work: https://github.com/binwiederhier/ntfy/commit/6d15b9face0f6dcc98ff2d0dfe95b47700139454

You could try it out :grimacing: 

 Are you sure that ALL the fields are necessary, or only the `poll_id`? I think that was missing.

barart:
@binwiederhier Nope, all fields are required iOS needs them (working on the fix i read that) is very strict specially on mute messages (this is the case), once you have time, read my previous post so you can understand, i already tried a lot of things to figure and protect the messages and that was my result of a lot of debug, what you did will leak the messages, not trigger the mute function (that was my first try) and some notifications will be lost

binwiederhier:
I wrote a big whole thing about how you are incorrect, and then I finally got it. My misunderstanding was that I was thinking that `data` was already censored, but we're in fact looking at `m`, which was not censored/converted to a poll request.

I'll update the code tonight to fix my fuckup.

binwiederhier:
You say that all fields are required for it to work in iOS, though this code does not have all the fields: https://github.com/binwiederhier/ntfy/blob/f40023aa232696637b0cd50d08b4f146091d0013/server/server_firebase.go#L130-L139

Can you explain why all the fields would be necessary? This^^ code would not work I think (this is for self-hosted folks), if all fields were necessary.

PR #1289, opened 2025-03-11
Add optional twilio-call-format config option

mmichaa:
To be able to set custom TwiML send to the Call API.


binwiederhier:
This is missing the `serve.go` config option, no?

binwiederhier:
Thanks. I've updated it a tad, so you can use Go templates in the call format. https://docs.ntfy.sh/config/#phone-calls

It'll be in the next server release.

PR #1296, opened 2025-03-19
Add Terminal Notifications for Long-Running Commands example

sharjeelaziz:
This is an example to add an alias to .bashrc and send notifications after long running commands complete. Thank you for creating ntfy.sh


binwiederhier:
Oh this is so cool!

PR #1299, opened 2025-03-25
docs: add integration: Monibot

cvilsmeier:
Monibot monitors your websites, servers and applications and notifies you if something goes wrong.
https://monibot.io/


cvilsmeier:
Dankeschön!

PR #1312, opened 2025-04-07
Consider aes128gcm content encoding as an indicator for UnifiedPush

vkrause:
Without this a UnifiedPush/Web Push message with encryption would be turned into an attachment. That in itself isn't pretty but can still work, but it requires attachments to be enabled in the first place.


zaq1758865:
> 

> 

Bash/sysOs-

Issue #1316, opened 2025-04-21
Publicly exposed NTFY web interface allowing unauthorized publish or subscribe access.


Remonsazzad:
ntfy-web-exposure
vulnerable ips : 

 http://147.135.1.155/settings
 http://147.135.114.248/settings
 http://159.203.157.154/settings
 http://135.148.243.12/settings
 http://142.93.11.244:8150/settings
 http://135.148.48.129/settings
 http://142.93.205.105/settings
 http://159.65.221.152/settings
 http://159.223.136.48/settings
 http://172.104.11.109/settings
http://143.42.6.46/settings
http://161.35.115.143/settings
http://172.232.10.86/settings
 http://172.232.9.32/settings
 http://172.232.5.155/settings
 http://172.232.25.41/settings
 http://172.232.31.94/settings
http://108.174.61.219:8080/settings
 http://172.232.11.95/settings
 http://129.159.101.178/settings
http://100.26.21.146/settings
 http://150.136.138.66/settings
http://107.174.26.14/settings
http://150.136.208.148/settings
http://150.136.161.59/settings
http://172.233.219.67/settings
http://172.233.234.203/settings
http://178.128.158.36/settings
http://173.255.230.134:8080/settings
http://172.233.223.21/settings
http://172.233.223.181:8092/settings
 http://172.234.195.69/settings
 http://172.234.207.20/settings
 http://172.233.221.122/settings
 http://172.234.26.66/settings
 http://206.189.181.180/settings
 http://178.156.149.97/settings
 http://3.82.114.241/settings
 http://173.211.12.34:8000/settings
 http://198.12.65.10:2222/settings
 http://192.3.24.21/settings
http://23.95.15.227:3051/settings
http://192.3.164.201:8001/settings
http://23.94.76.46:8080/settings
http://191.96.165.77:8150/settings
http://193.122.140.222/settings
http://192.210.197.55:2200/settings
http://45.79.141.103/settings
 http://45.79.151.15/settings
http://45.33.85.167/settings
 http://44.219.179.106/settings
http://45.79.151.136/settings
http://64.227.12.13/settings
http://50.116.54.38/settings
 http://64.227.16.41:5005/settings
http://89.116.44.198/settings
http://67.205.164.177:8080/settings
http://66.228.41.26/settings
 http://54.237.5.168/settings
https://107.172.88.145/settings
https://107.173.144.88/settings
 https://135.148.33.129/settings
https://159.223.136.48/settings
[...]

PR #1329, opened 2025-05-05
[Featute] Add optional web app flag which requires a login for every action

theatischbein:
## Problem
I have read #238 and was also notified about #1316 (which I disagree too because the ACL does not allow it).
I do understand that the web app is supposed to be a dump client, but I still want to manage who uses my server ressources. Since the web app has a login it feels not suitable to add an additonal login step like BASIC-AUTH.

But I see the point of #238 that the web app can be used to subscribe to other ntfy servers, which might not be wanted.

Also I find it a bit confusing in my setup that there are options in the NavigationBar to subscribe or publish notifications on my servers without a login, but the ACL does not allow it. And I do not want others (without an account) to use my instance to subscribe to external server.

## Changes
So I added the flag `require_login`/`require-login`/`NTFY_REQUIRE_LOGIN` (Default: `false`).

If the flag is enabled and there is no Login:
- NavigationBar: hides All Notifications, Settings, Publisish Notification, Subscribe to topic
- NavigationBar: Documentation and all other possible entries are still there
- Endpoint `/settings` redirects back to main page
- Notification: Informs that a Login is needed to use this page

Screenshot from 2025-05-05 10-54-35

## Missing
I only added the translation for English.


binwiederhier:
So I fixed the missing pieces (see https://github.com/binwiederhier/ntfy/pull/1351). I tried it and I am a big fan of the `require-login` flag, though I am not fan of where the "this page requires a login" page is hooked into. I'd much rather it'd be instead of `<Main>` or even instead of `<Layout>`, or that we otherwise dynamically route to a different page. I would likely not want the nav/sidebar either.

I was going to get this into the next release, but it may have to wait, since there is much more to look at.

binwiederhier:
@theatischbein Did you want to take another look at this^? Otherwise I think I'll have to redo/close it

theatischbein:
thanks you very much! :)

tim-koehler:
Hey @binwiederhier 
when can we expect this to be officially released? 😊

PR #1328, opened 2025-05-05
Add ability to build the server for Windows and run it as a service

wtf911:

PR #1355, opened 2025-05-27
Update install.md - include Windows server

wtf911:
Add notes on running the server under Windows - related to #1328

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -223,11 +223,15 @@ brew install ntfy
 
 
 ## Windows
-The [ntfy CLI](subscribe/cli.md) (`ntfy publish` and `ntfy subscribe` only) is supported on Windows as well.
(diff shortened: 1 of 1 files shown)


minhvn09:
> Add note về việc chạy máy chủ trên Windows - liên kết đến #1328

Can you give me a video tutorial on how to install a server for Windows? I'm not good at programming. Thank you very much.

wtf911:
> > Add note về việc chạy máy chủ trên Windows - liên kết đến #1328
> 
> Can you give me a video tutorial on how to install a server for Windows? I'm not good at programming. Thank you very much.

There's not currently an official release of the Windows server version, but here is my compiled version (use at your own risk obviously). It works the same as it does under any other operating system as far as the options to run it such as the --config flag to use a server.yml file or other options. It doesn't install like you think of a typical Windows program but it can be added as a Windows service. It runs from command line. I attached a newly compiled version that includes all of the latest commits.
[ntfy.zip](https://github.com/user-attachments/files/20555751/ntfy.zip)

PR #1371, opened 2025-06-19
Smaller docs updates

cyb3rko:
Hello there,

I've made some smaller changes to the docs:

- updated App store badges
- removed deprecated version field from Docker compose files


binwiederhier:
Curious: Why do we need new app badges?

cyb3rko:
Officially you should never use outdated Google badges: https://partnermarketinghub.withgoogle.com/brands/google-play/visual-identity/badge-guidelines ("Things to avoid")

The other ones were just low res

PR #1382, opened 2025-07-06
Add piping support

srevn:
The client will automatically detect piped input and use it as the message content when no explicit message is provided.

- Added stdin detection: The isStdinRedirected() function detects when data is being piped by checking if stdin is not a character device
- Automatic reading: When no message is provided via arguments and stdin has data, it automatically reads the piped content
- Preserves existing behavior: Still works with explicit `--file=-` and doesn't break any existing functionality

Example : `echo "test message from stdin" | ntfy publish mytopic`

### DIFF
--- a/cmd/publish.go
+++ b/cmd/publish.go
@@ -254,6 +254,16 @@ func parseTopicMessageCommand(c *cli.Context) (topic string, message string, com
 	if c.String("message") != "" {
 		message = c.String("message")
 	}
+	
+	// If no message provided and stdin has data, read from stdin
+	if message == "" && stdinHasData() {
+		var stdinBytes []byte
+		stdinBytes, err = io.ReadAll(c.App.Reader)
+		if err != nil {
+			return
+		}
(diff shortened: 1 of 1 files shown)


cmd/publish.go
```diff
@@ -254,6 +255,15 @@ func parseTopicMessageCommand(c *cli.Context) (topic string, message string, com
 	if c.String("message") != "" {
 		message = c.String("message")
 	}
+	if message == "" && isStdinRedirected() {
+		var bytes []byte
+		bytes, err = io.ReadAll(c.App.Reader)
```
binwiederhier, line 260: We don't ever want to do unbounded reads into memory. This will OOM. Limit to a MB or something. I think wrapping this in a limit reader will work

srevn: Ok, noted, thanks! Anything else?

binwiederhier: I'm on my phone, I'll look later.

binwiederhier:
Works like a charm. Thanks

PR #1384, opened 2025-07-07
Declarative users

binwiederhier:
This implements declarative users:

```yaml
auth-users:
  - "phil:$2a$10$YLiO8U21sX1uhZamTLJXHuxgVC0Z/GKISibrKCLohPgtG7yIxSk4C:admin"
  - "ben:$2a$10$NKbrNb7HPMjtQXWJ0f1pouw03LDLT/WzlO9VAv44x84bRCkh19h6m:user"
auth-access:
  - "ben:alerts-*:rw"
  - "ben:system-logs:ro"
  - "*:announcements:ro" # or: "everyone:announcements,ro"
auth-tokens:
  - "phil:tk_3gd7d2yftt4b8ixyfe9mnmro88o76"
  - "ben:tk_f099we8uzj7xi5qshzajwp6jffvkz:Ben personal token"
```


user/manager_test.go
```diff
@@ -1075,6 +1096,24 @@ func TestManager_Topic_Wildcard_With_Underscore(t *testing.T) {
 	require.Equal(t, ErrUnauthorized, a.Authorize(nil, "mytopicX", PermissionWrite))
 }
 
+func TestManager_WithProvisionedUsers(t *testing.T) {
+	f := filepath.Join(t.TempDir(), "user.db")
+	conf := &Config{
+		Filename:      f,
+		DefaultAccess: PermissionReadWrite,
+		ProvisionedUsers: []*User{
+			{Name: "phil", Hash: "$2a$10$YLiO8U21sX1uhZamTLJXHuxgVC0Z/GKISibrKCLohPgtG7yIxSk4C", Role: RoleAdmin},
```
wunter8, line 1105: It doesn't matter if the bcrypt cost of a particular provisioned user's password hash (e.g., `$2a$XX$...`) is different from the server's configured `BcryptCost`, right?

I haven't looked close enough at the code or tested it myself, but it seems like it could be possible that the server is configured to use a cost of `10` and the provisioned hash used a cost of `12`. Then, when trying to login to the provisioned user's account, the server will generate `hash(<password>, cost: 10)` and compare it to the hash in the DB, but they wouldn't match

user/manager.go
```diff
@@ -427,6 +431,15 @@ const (
 	migrate4To5UpdateQueries = `
 		UPDATE user_access SET topic = REPLACE(topic, '_', '\_');
 	`
+
+	// 5 -> 6
+	migrate5To6UpdateQueries = `
+		ALTER TABLE user ADD COLUMN provisioned INT NOT NULL DEFAULT (0);
+		ALTER TABLE user ALTER COLUMN provisioned DROP DEFAULT;
```
wunter8, line 438: I haven't seen this before, where a column is added with a default and then the default is immediately removed. Is this just a sqlite quirk?

user/manager.go
```diff
@@ -1505,6 +1578,67 @@ func (a *Manager) Close() error {
...
+			}
+			existingUser, exists := util.Find(users, func(u *User) bool {
+				return u.Name == user.Name
+			})
+			if !exists {
+				log.Tag(tag).Info("Adding provisioned user %s", user.Name)
+				if err := a.addUserTx(tx, user.Name, user.Hash, user.Role, true, true); err != nil && !errors.Is(err, ErrUserExists) {
+					return fmt.Errorf("failed to add provisioned user %s: %v", user.Name, err)
+				}
+			} else if existingUser.Hash != user.Hash || existingUser.Role != user.Role {
```
wunter8, line 1617: You're searching through all existing users in the DB, not just provisioned users. So it seems like this could change the password or role of a normal (non-provisioned) user. Is that intended?

wunter8:
You mentioned on Discord that the YAML file is set up the way it is to support ENV variables. However, I didn't see ENV variables being used anywhere

user/manager.go
binwiederhier: Im gonna take a look. Good catch!

user/manager_test.go
binwiederhier: Good catch. I'll go take a look.

binwiederhier:
In serve.go, all the comma d line args are also configurable via env variables. The cli lib does that for us. That's why we can't do nested yaml. A list is the best we can do.

cmd/serve.go
```diff
@@ -517,6 +531,76 @@ func parseIPHostPrefix(host string) (prefixes []netip.Prefix, err error) {
...
+		if username == userEveryone {
+			username = user.Everyone
+		}
+		provisionUser, exists := util.Find(provisionUsers, func(u *user.User) bool {
+			return u.Name == username
+		})
+		if username != user.Everyone {
+			if !exists {
+				return nil, fmt.Errorf("invalid auth-provision-access: %s, user %s is not provisioned", accessLine, username)
+			} else if !user.AllowedUsername(username) {
```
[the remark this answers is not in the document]
binwiederhier: Yeah, cheap validation doesn't hurt to do in more than one place.

PR #1387, opened 2025-07-13
Fixed broken codeblocks in Markdown config.md

watou:
Also improve syntax highlighting where possible.


binwiederhier:
Thank you for your contribution, however, the docs you "fixed" are not broken. We are using MkDocs for Material to render enhanced Markdown, see https://docs.ntfy.sh/config/ for the rendered version

watou:
There was the hope that a single .md file could be rendered properly on both Github Markdown and MkDocs, but as you like.

PR #1392, opened 2025-07-17
FEAT : add get topic subscribers endpoint

MohamedElbashar:
#  ntfy Subscriber Endpoint Implementation

##  Overview

This implementation adds a new REST API endpoint to ntfy that allows you to check if a topic has active subscribers and get real-time connectivity information between topics and consumers.

###  New Feature Added

**Endpoint:** `GET /v1/topics/{topic}/subscribers`

This endpoint returns real-time information about whether anyone is currently subscribed to a specific topic, enabling you to verify connectivity before sending important notifications.

### Check if anyone is subscribed to a topic:
```bash
curl http://localhost:8080/v1/topics/my-topic/subscribers
```

### Response format:
```json
{
  "topic": "my-topic",
  "subscribed": true,
  "count": 2,
  "last_access": "2025-07-17T13:30:17Z"
}
```

---

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -87,6 +87,7 @@ var (
 	metricsPath                                          = "/metrics"
 	apiHealthPath                                        = "/v1/health"
 	apiStatsPath                                         = "/v1/stats"
+	apiTopicSubscribersPath                              = "/v1/topics/%s/subscribers"
 	apiWebPushPath                                       = "/v1/webpush"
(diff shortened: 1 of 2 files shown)


binwiederhier:
Thank you for your contribution, but I don't think I want a subscriber endpoint like this in the API. If you are sending sensitive stuff, you should protect topics with an ACL entry.

PR #1395, opened 2025-07-20
doc corrections

wunter8:


wunter8:
The page with all the template function descriptions (`/publish/template-functions`) isn't in the navigation sidebar at all.

Right now, the only way to get to it is by selecting one of the function type headings under `/publish/#template-functions`

Is that intentional?

binwiederhier:
Awesome thank you!

binwiederhier:
Yes, it is intentional that the template functions are not in the nav bar. Mainly because adding it creates the most outrageous looking sidebar ever. But I think I'll do it anyway. You've got a point

PR #1397, opened 2025-07-22
Add a lock to add_messages

### LINKED ISSUE #1391: http 500 error (ntfy error 50001)
When sending a notification (especially when it happens with multiple requests) this error occurs

ntfy server 2.13.0 in docker, debian 12 arm64

```
closed with HTTP 500 (ntfy error 50001) (error=database table is locked, http_method=POST, http_path=/_matrix/push/v1/notify, tag=http, visitor_auth_limiter_limit=0.016666666666666666, visitor_auth_limiter_tokens=30, visitor_id=ip:<edited>, visitor_ip=<edited>, visitor_messages=448, visitor_messages_limit=17280, visitor_messages_remaining=16832, visitor_request_limiter_limit=0.2, visitor_request_limiter_tokens=57.049697891799994, visitor_seen=2025-07-16T15:06:35.429Z)
```

Looks like this has already been fixed by #498, regression?

timofej673:
Fixes #1391
According to the rules of sqlite - there can only be one writer
So my change adds a lock via a mutex to add_messages to comply with this rule and not cause error "database is locked". I tested my version - everything works for me and this error disappeared
Also I formatted the code a little so that there are no extra tabs and spaces and something like that


binwiederhier:
This is quite odd and unnecessary if you run sqlite in WAL mode and enable batching, you should not need this. Have you followed https://docs.ntfy.sh/config/#message-cache_1?

I am inclined to still merge this PR, because it should not be harmful at all, even for batching.

timofej673:
> This is quite odd and unnecessary if you run sqlite in WAL mode and enable batching, you should not need this. Have you followed https://docs.ntfy.sh/config/#message-cache_1?
> 
> I am inclined to still merge this PR, because it should not be harmful at all, even for batching.

Hello. Yes, I enabled WAL mode, set busy_timeout, enabled batch mode. And everything else that was in the documentation. However, the errors did not disappear

I even increased busy_timeout to 5 minutes. But the errors still did not disappear

I even set journal_size_limit to 1 megabyte. Still did not help

And increased "cache-batch-timeout" to 10 seconds. Same.

I usually do not like to dig into the code, besides, I do not know go language very well. However, I had to do this, since the documentation and other methods did not help me.

And I began to read more about busy_timeout and found that it is suitable for cases when there are several connections to one database. However, in the code I found that only one connection is created to the message cache database. And several goroutines access it. That is, simultaneous write access to one connection. And if the error was due to low busy_timeout, I would get "database is busy", but this is obviously not my case.

And I also found out why WAL didn't help. Because the rule of one writer still works there! The documentation says: "there can only be one writer at a time" https://www.sqlite.org/wal.html. Yes, it now allows multiple readers to read while the writer is working, **however, there can still only be one writer**. With multiple writers, that same "database is locked" error occurs

Well, based on this, I came to the conclusion that the only way to fix "database is locked" is to set a mutex on the writer function. I made this commit, recompiled the docker image and ran it on my machine. The error completely disappeared. At least this program with my patch has been working for me for 3 days now.

I would be very glad if you accept this pull request, this is my first one :)

binwiederhier:
> And I began to read more about busy_timeout and found that it is suitable for cases when there are several connections to one database. However, in the code I found that only one connection is created to the message cache database. And several goroutines access it. That is, simultaneous write access to one connection. And if the error was due to low busy_timeout, I would get "database is busy", but this is obviously not my case.

I found this out last night too, and I started dabbling a lot: https://github.com/binwiederhier/ntfy/pull/1405/commits/f4c285f7ce722c37a1f12a5dfdd4f212385979bf

What I want to do is this: 
- [ ] Make the default busy_timeout=5000 by modifying the DSN from just the filename to `file:$file?_busy_timeout=5000`. See https://github.com/mattn/go-sqlite3?tab=readme-ov-file#connection-string
- [ ] Make the DSN configurable in the config file, e.g. as `auth-file: "file:/tmp/cache.db?_busy_timeout=90000"` or `auth-file: /tmp/cache.db` (defaults to 5s busy timeout).
- [ ] Update the docs

I don't think the lock would be necessary in this case, since SQLite would do the locking/queuing.

Your code works and is a fine solution. I may still take the PR, but I still want to do the configurable busy timeout and the default 5s timeout.

binwiederhier:
I tried the busy timeout stuff, and it is not worth it. I'll do your approach, likely this weekend. I'll pull in your changes for that. It worked great in my testing.

binwiederhier:
Thanks @timofej673

PR #1404, opened 2025-07-27
allow newlines in in-line go templates

wunter8:
Before the changes in this PR, there was a difference between
```
message: {{.foo}}{{"\n"}}{{.bar}}
```
in a template YAML file and
```
m={{.foo}}{{"\n"}}{{.bar}}
```
in an inline template (in a URL/header param).

The inline template version would break (be considered an invalid template), because we were replacing all "\n" in the message field with literal newlines, making the template look like this:
```
{{.foo}}{{"
"}}{{.bar}}
```

Doing that before we rendered the template caused problems. This changes the code so that when an inline template is being used, we don't replace "\n" with literal newlines until after the template is rendered.


binwiederhier:
Thanks!

PR #1406, opened 2025-07-30
refactor: use slices.Contains to simplify code

tanhuaan:
There is a [new function](https://pkg.go.dev/slices@go1.21.0#Contains) added in the go1.21 standard library, which can make the code more concise and easy to read.


binwiederhier:
Thanks

PR #1425, opened 2025-08-17
feat(docs): add ntfy-heartbeat-monitor to integrations page

DerRockWolf:
I've written a small tool to implement heartbeat monitoring/alerting.
My main use-case is to be alerted if something in my alerting pipeline is broken.

It works by monitoring messages in a ntfy topic. If the last message is older than a given threshold, or if we can't retrieve the last message because the ntfy server is down/unavailable, a message is sent to a different ntfy topic (usually also on a different ntfy server).
If you have two or more "sites" you could monitor everything in a mesh like way :smile:


binwiederhier:
Love it

PR #1430, opened 2025-08-23
Fix base64 snippets in Publishing

hxtmdev:
`-w0` is usually needed for longer outputs

Might be GNU-specific, though.

To quote the man page

>        -w, --wrap=COLS
>               wrap encoded lines after COLS character (default 76).  Use 0 to disable line wrapping


binwiederhier:
Thanks

PR #1433, opened 2025-08-24
convert old for loops to range functions

JoshGuarino:
Since this project tries to use the latest go version,  I thought it prudent to update some the old for loops what can be converted to using range function.  Language server was giving warnings `for loop can be modernized using range over int`  so I figured I would do some of this cleanup work.


binwiederhier:
Thank you for your contribution. 

I am not really a fan of this syntax. It's syntax sugar that makes things harder to read IMHO, so I am closing this.

Issue #1452, opened 2025-09-30
Docs: Document topic length requirements and add generator


rugk:
:bulb: **Idea**

Docs page I talk about: https://docs.ntfy.sh/publish/

I've split the doc ideas from https://github.com/binwiederhier/ntfy/issues/1451, these are:

* "the topic is essentially a password" should IMHO be a bigger warning (like [GitHub's markdown warning block](https://github.com/orgs/community/discussions/16925) as I had to read the page a little more than needed to find that doc snippet. And it's important, because if this gets more popular I'd bet many people – in the worst case – use the default string mentioned in the docs .
* (added) Add a note about them maximum length of topics!
* to ease the UX even more and convince people this (= security) is actually a good idea, you cold even add a little generator for your topic URL that locally i the client (via JS), generates a topic like `topicname_somelonghash`. Have a big button "generate topic name" and I guess this will easily lead to people using it.  
   Also, this makes it easier for new users like me that first have to understand the doc what a "topic" actually is/refers to. As I came from that third-party application linking to you with a text like this, it may be a good idea to have it:
   > For ntfy, you need to create a topic on your preferred ntfy instance. This is pretty easy. Please visit the steps in the [docs](https://docs.ntfy.sh/publish/) and follow the instructions.

   If you want to make it even better, you could create a separate page or even workflow for a "generate URL and return to third-party" (like OAuth) that simplifies the UX to simply pressing button to generate that and return it to the client. 

:computer: **Target components**

web (but docs, not the web app!)

binwiederhier:
Did this. Will release it soon. 

<img width="1578" height="675" alt="Image" src="https://github.com/user-attachments/assets/5e5e7a3b-aaa5-41b5-acb4-693685903a20" />

PR #1453, opened 2025-09-30
Add warning about topic length and secrecy note

rugk:
This addresses the two simple concerns in https://github.com/binwiederhier/ntfy/issues/1452 regarding documentation. Though I just applied GitHub's markdown syntax here, I have no idea whether that works for you too. (What doc generator is even used here? Found no information in docs folder/contrib guide etc.?)

### DIFF
--- a/docs/publish.md
+++ b/docs/publish.md
@@ -1,7 +1,13 @@
 # Publishing
 Publishing messages can be done via HTTP PUT/POST or via the [ntfy CLI](install.md). Topics are created on the fly by 
-subscribing or publishing to them. Because there is no sign-up, **the topic is essentially a password**, so pick 
-something that's not easily guessable.
+subscribing or publishing to them. 
+
+> [!IMPORTANT]  
+> Because there is no sign-up, **the topic is essentially a password**, so pick 
+something unique that's not easily guessable.
+ 
+> [!NOTE]  
+> Note a topics length is limited technically to 64 characters.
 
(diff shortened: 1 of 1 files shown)


rugk:
Ah found another docs page I copied the syntax from.

docs/publish.md
```diff
@@ -1,7 +1,13 @@
...
-subscribing or publishing to them. Because there is no sign-up, **the topic is essentially a password**, so pick 
-something that's not easily guessable.
+subscribing or publishing to them. 
+
+!!! warning
+    Because there is no sign-up, **the topic is essentially a password**, so pick 
+something unique that's not easily guessable.
+ 
+!!! info
+    Note a topics length is limited technically to 64 characters.
```
wunter8, line 10: Typo: topic's length

rugk:
```suggestion
    Note a topic's length is limited technically to 64 characters.
```

you mean?

binwiederhier:
Honestly, I think the bold text is enough. It's written everywhere all over the docs. I don't see the reason for a scary warning like that. 

Thank you for your contribution, but I think I'd like to reject your PR

rugk:
I've only seen it written one time on the docs page, so I must have missed it being written somewhere else. Anyway, what about including this part then at least, which is clearly not written anywhere?
> Note a topic's length is limited technically to 64 characters.

PR #1466, opened 2025-10-18
WIP: Update notifications

wunter8:
This allows notification content to be updated in the web app by sending multiple messages with the same `sequence id` (`sid`).

Here is a small video demo:

https://github.com/user-attachments/assets/bc020d1e-5772-446e-bf97-dd6759a0dddc

I tried to make sure that things still work if this new version of the web app subscribes to a topic on an "old" ntfy server (a server that doesn't provide `mtime` or `sid` fields).

The server doesn't support "deleting" notifications (dismissing from the notification shade) yet, just updating.

The web app cannot "delete" notifications programmatically since "silent" pushes are not allowed. (i.e., sending a JSON object with a message ID to delete and not showing any notification as a result of that JSON object is not allowed. Deleting should work fine on Android, though.)

@binwiederhier you'll need to decide how we present all this visually in the web app. I just wanted to get the information on there somewhere.

All the revisions of a particular message are attached to the one visible notification, so we could create some sort of dialog/modal to show the past revisions of a message. But I figured we could do that later, once everything else is working.

### DIFF
--- a/web/public/static/langs/en.json
+++ b/web/public/static/langs/en.json
@@ -70,6 +70,8 @@
   "notifications_delete": "Delete",
   "notifications_copied_to_clipboard": "Copied to clipboard",
   "notifications_tags": "Tags",
+  "notifications_sid": "Sequence ID",
+  "notifications_revisions": "Revisions",
   "notifications_priority_x": "Priority {{priority}}",
   "notifications_new_indicator": "New notification",
   "notifications_attachment_image": "Attachment image",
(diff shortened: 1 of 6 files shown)


wcypierre:
Any updates on this?

wunter8:
Not yet, no

binwiederhier:
Superseded by https://github.com/binwiederhier/ntfy/pull/1536

PR #1467, opened 2025-10-19
Add sandboxing to ntfy.service

Velocifyer:
See [systemd.exec(5)](https://man.archlinux.org/man/systemd.exec.5) to find out what the options mean!

### DIFF
--- a/server/ntfy.service
+++ b/server/ntfy.service
@@ -10,6 +10,18 @@ ExecReload=/bin/kill --signal HUP $MAINPID
 Restart=on-failure
 AmbientCapabilities=CAP_NET_BIND_SERVICE
 LimitNOFILE=10000
+PrivateDevices=true
+ProtectClock=true
+ProtectKernelTunables=true
+ProtectKernelModules=true
+ProtectKernelLogs=true
+RestrictRealtime=true
+ProtectHostname=true
+
+# These will be added in a future update.
+# ProtectSystem=full
+# PrivateTmp=true
+
 
 [Install]
 WantedBy=multi-user.target


Velocifyer:
Ideally `ProtectSystem=strict` would be used along with `ReadWritePaths=` and `ReadOnlyPaths=`, but I don't know every path ntfy uses.

binwiederhier:
I tried this out and most seem harmless. However, `PrivateTmp=true` and `ProtectSystem=full` may cause problems with people's installations:

- PrivateTmp breaks how ntfy is installed on ntfy.sh, because it uses /tmp as the socket and nginx can't see it
- ProtectSystem=full may cause problems if people put their sqlite databases into /etc (which they obviously shouldn't)

So I think I'll add the others and update the docs with hardening suggestions. Sound good?

Velocifyer:
>  PrivateTmp breaks how ntfy is installed on ntfy.sh, because it uses /tmp as the socket and nginx can't see it

You should be using /run.

>ProtectSystem=full may cause problems if people put their sqlite databases into /etc

databases should not be in /etc.

If you want i can ammend the commit to remove PrivateTmp. If a lot of people are putting sockets in /run i reccommend that you still add PrivateTmp but put a warning in the changelog.

Velocifyer:
Do you want any changes? Should I add RestrictSUIDSGID and ProtectHostname and ProtectHome and ProtectControlGroups?

binwiederhier:
We can keep the "harmless ones" and get those merged, and then put up a notice on the deprecation warning page that we'll make the systemd service security tighter, but I think given that we have tens of thousands of installations probably, I don't want to break people's stuff for this.

Velocifyer:
Is this better?

binwiederhier:
Thanks

omegasupreme3:
Thought I would just add my experience with _systemd_ hardening in relation to ntfy.

I am running RHEL 10, minimal server installation. Deployed `ntfy_2.22.0_linux_amd64.rpm` and just kept it to the default configuration (e.g. SQLlite DB). After that I applied the following _systemd_ constraints:

[Service]
_**NoNewPrivileges=yes
ProtectSystem=strict**_
ProtectKernelTunables=yes
ProtectKernelModules=yes
ProtectControlGroups=yes
_**RestrictNamespaces=yes**_
RestrictRealtime=yes
_**RestrictSUIDSGID=yes
LockPersonality=yes
InaccessiblePaths=/root /srv
ReadOnlyPaths=/etc
ReadWritePaths=/etc/ntfy/ /var/lib/ntfy /var/cache/ntfy**_
ProtectClock=yes
ProtectKernelLogs=yes
_**MemoryDenyWriteExecute=yes
ProtectHostname=yes
ProtectProc=ptraceable**_
_**SystemCallFilter=~@reboot @resources @clock @debug @obsolete @module @pkey
CapabilityBoundingSet=~CAP_KILL CAP_LINUX_IMMUTABLE CAP_SYS_BOOT CAP_SYS_ADMIN
ProtectHome=yes
PrivateUsers=yes**_
PrivateDevices=yes

I totally understand you're careful with tightening the security as it may easily prohibit ntfy from working correctly. Especially as systems are hardly the same, so approach with care. Nonetheless I would like to mention that in my case the above measures are working flawlessly for over a month now. I marked the extra measures I did bold.

Checkout this link for more information: [https://linux-audit.com/systemd/settings/units/](url).

Just my 2 cents. ;-)

Issue #1423, opened 2025-08-10
Register the ntfy PWA as a share target


zamuz:
:bulb: **Idea**
It would be useful for users of the ntfy web app to be able to share from the Android Share menu directly to the ntfy web app. This should be possible via the Web Share Target API:

https://developer.chrome.com/docs/capabilities/web-apis/web-share-target

:computer: **Target components**
Web app, maybe configurable in the server configuration.

Thank you every much in advance!

binwiederhier:
Sounds like a good idea. Do it !😂

PR #1495, opened 2025-11-22
enable pwa share target

nexus-uw:
- adds share_target property to manifest with title + text + url
- configure to send POST to new url , /share-target
  - going with POST so that files can be shared in the future -> https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Manifest/Reference/share_target#receiving_shared_files
- updated webworker to intercept fetch requests, and redirect+restructure /share-target requests to SPA page with parameters moved to the url params

 (mostly generated by gpt5 using
"i want to add a pwa share target that invokes the publish message button on the web app")

note: I am unable to currently build the docker container to fully test the e2e flow due to 

````
25.40 cmd/serve.go:22:2: no required module provides package heckel.io/ntfy/v2/payments; to add it:
25.40   go get heckel.io/ntfy/v2/payments
25.42 make: *** [Makefile:188: cli-linux-server] Error 1
------

 1 warning found (use docker --debug to expand):
 - FromAsCasing: 'as' and 'FROM' keywords' casing do not match (line 1)
Dockerfile-build:43
--------------------
  41 |     ADD ./user ./user
  42 |     ADD ./util ./util
  43 | >>> RUN make VERSION=$VERSION COMMIT=$COMMIT cli-linux-server
````

running the web client with the npm start command, I was able to verify that url params are successfully read from the url and the publish message opened with the expected values.

what would be the best test command(s) to get the server + web assets deployed locally together?

implements https://github.com/binwiederhier/ntfy/issues/1423

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 MAKEFLAGS := --jobs=1
 PYTHON := python3
 PIP := pip3
-VERSION := $(shell git describe --tag)
+VERSION := 1111
 COMMIT := $(shell git rev-parse --short HEAD)
 
 .PHONY:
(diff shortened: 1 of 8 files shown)


binwiederhier:
I am happy to accept PRs, even if AI was used, but not if they are not if I have to do all the work of testing and fixing.

PR #1499, opened 2025-11-25
Add Simple Observability to integrations list

khazit:
Simple Observability is a server monitoring and observability platform that supports using ntfy as a notification channel for alerting


binwiederhier:
Very cool. Thanks.

PR #1501, opened 2025-11-28
do not build fbsend with nofirebase

aerusso:
The nofirebase build tag should remove all build dependencies on firebase.  The fbsend test tool depends on firebase (and is also only useful in builds that use firebase).  Hence, disable it.


binwiederhier:
Thanks

PR #1512, opened 2025-12-07
do not escape JSON response

### LINKED ISSUE #1511: Ampersand character `&` is replaced with `\u0026`
:question: **Question**

I am sending this URL `https://drive.google.com/thumbnail?id=1kCU5ThTEnbk3ltqs_TFBZVKUFYiZ3l5C&sz=w1000` but my `journalctl` reports this final output
```
{"id":"CmxIRVnmeowd","time":1765128847,"expires":1765172047,"event":"message","topic":"hometr","message":"Human detected","tags":["rotating_light"],"attachment":{"name":"thumbnail","url":"https://drive.google.com/thumbnail?id=1QLk2zPn27AMJdbg2KxQ0RBVUx0dPrUuh\u0026sz=w1000"}}
```

As you can see the character `&` is replaced with `\u0026` 

How can I avoid this?

wunter8:
Apparently, golang's JSON encoder automatically escapes `<`, `&`, and `>` in a JSON response (See https://github.com/golang/go/issues/28453 and https://pkg.go.dev/encoding/json#Encoder.SetEscapeHTML).

This was unexpected to me. It seems like whoever is using the returned ntfy message should choose to escape the data if they want to before they use it, instead of ntfy always escaping the data.

Before the change, `curl -d '&' https://ntfy.sh/test` would return `{..."message":"\u0026"...}`

After the change, `curl -d '&' https://ntfy.sh/test` returns `{..."message":"&"...}`

Fixes #1511

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -1271,7 +1271,9 @@ func (s *Server) handleBodyAsAttachment(r *http.Request, v *visitor, m *message,
 func (s *Server) handleSubscribeJSON(w http.ResponseWriter, r *http.Request, v *visitor) error {
 	encoder := func(msg *message) (string, error) {
 		var buf bytes.Buffer
-		if err := json.NewEncoder(&buf).Encode(&msg); err != nil {
+		encoder := json.NewEncoder(&buf)
+		encoder.SetEscapeHTML(false)
+		if err := encoder.Encode(&msg); err != nil {
 			return "", err
 		}
 		return buf.String(), nil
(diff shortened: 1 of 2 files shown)


binwiederhier:
That seems very straight forward, though I wonder what that will break downstream. This feels like a "Hyrum's Law" thing.

wunter8:
Possibly. We could put it behind a header/query param flag then? Something like `escapeHtml=false`?

binwiederhier:
I'm wondering if we should just merge this and just stick a note in the release notes. It does feel odd to HTML encode in JSON responses..

binwiederhier:
What do you think @wunter8 ?

wunter8:
I agree that it feels odd to encode the JSON body. I think we're probably okay releasing this with a note

PR #1531, opened 2026-01-02
Added optional TLS encryption for SMTP server

### LINKED ISSUE #409: Add TLS encryption to SMTP receiver
As of today the e-mail receiving server (for e-mail publishing) is only able to accept plain text connections.
I'd like to be able to use TLS encrypted (STARTTLS) connections, as notification contents might contain sensitive information.
As an option, I'd like to be able to make TLS mandatory on the server side, i.e. reject not encrypted connections via an config option in the server.yml.

santanoce:
This would close #409.

I added the following configuration options:
- `smtp-server-cert-file` to specify the public key file;
- `smtp_server_key_file` to specify the private key file;
- `smtp-server-implicit-tls` to force TLS (Implicit TLS, otherwise it will use STARTTLS).

If the above options are not specified, the SMTP server would just accept plain text connections, as it is now.


docs/config.md
```diff
@@ -619,13 +619,16 @@ users can [send emails to a topic e-mail address](publish.md#e-mail-publishing)
 `myprefix-mytopic@ntfy.sh`) to publish messages to a topic. This is useful for e-mail based integrations such as for 
 statuspage.io (though these days most services also support webhooks and HTTP calls).
 
-To configure the SMTP server, you must at least set `smtp-server-listen` and `smtp-server-domain`:
+To configure the SMTP server, you must at least set `smtp-server-listen` and `smtp-server-domain` (TLS can be optionally be enabled providing the required certificate files):
```
wunter8: Typo

PR #1535, opened 2026-01-05
Fix curl command example in Notifications component

jjasghar:
It seems at least in my testing, no response happens without the `https://` prepended.

for instance:
```
➜  ~ curl -d 'hi' ntfy.example.com/p6r76PFAKE_THINGHtRZ3x
➜  ~ curl -d 'hi' https://ntfy.example.com/p6r76PFAKE_THINGHtRZ3x
{"id":"LURnWXPPmkDM","time":1767634827,"expires":1767678027,"event":"message","topic":"p6r76PFAKE_THINGHtRZ3x","message":"hi"}
```

### DIFF
--- a/web/src/components/Notifications.jsx
+++ b/web/src/components/Notifications.jsx
@@ -600,7 +600,7 @@ const NoNotifications = (props) => {
       <Paragraph>
         {t("notifications_example")}:<br />
         <tt>
-          {'$ curl -d "Hi" '}
+          {'$ curl -d "Hi" https://'}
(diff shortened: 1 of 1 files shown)


epifeny:
IMO, one can (preferably for internal uses only), use http and not always https, and so `curl -k` would still work.

binwiederhier:
Fixed in https://github.com/binwiederhier/ntfy/commit/946a2b6fbe3b512de84b43e544f25a7aa429f040

PR #1551, opened 2026-01-17
Update link to ntfysh-windows client

mshafer1:
Since lucas-bortoli marked the original project as archived/abandoned, I intend to continue maintenance on my fork. Therefore, updating the integration link to point to it.

CC @alexhorner as other contributor to original project to confirm that we want mshafer1/ntfysh-windows to become the official fork.


binwiederhier:
Thanks

PR #1568, opened 2026-01-24
Fix links in contributing.md for development guide and contact to work from project-root-page

BeeJay28:
When you click on the links from the Contribution.md file, they work just fine.
If you click on the links from the project-root page, it directs you to a 404-page, since the links are relative.

Fix by making link-URIs absolute (relative to the project-root, I guess that's how Github works)

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -7,9 +7,9 @@ translator, or just an enthusiastic user.
 
 If you'd like to contribute code to ntfy:
 
-1. Check out the [development guide](develop.md) to set up your environment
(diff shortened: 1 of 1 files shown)


binwiederhier:
Thanks for trying to help, but the docs are not raw markdown, they are rendered by mkdocs, see https://docs.ntfy.sh/contributing/ -- So the links work just fine :-D

BeeJay28:
I see. They don't work via the Github-Repository's root-page, but I guess the ntfy.sh page outweighs here :D

PR #1577, opened 2026-01-27
[Docs/publish] Fix JSON syntax for JS examples

KavyanshKhaitan2:
(and make one of them more consistent)

### DIFF
--- a/docs/publish.md
+++ b/docs/publish.md
@@ -1619,7 +1619,7 @@ And the same example using [JSON publishing](#publish-as-json):
         method: 'POST',
         body: JSON.stringify({
             topic: "myhome",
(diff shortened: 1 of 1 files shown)


KavyanshKhaitan2:
sorry didnt notice the other errors while making the last PR.
noticed it now when checking if it got deployed, and saw that the same mistake is in other places too.
also the change i made wasnt consistent with the other KV pairs, so made that more consistent for the docs.

KavyanshKhaitan2:
that was fast...

KavyanshKhaitan2:
oh well!

KavyanshKhaitan2:
@binwiederhier btw just curious but do you manually deploy it..?
how do you deploy ntfy after a commit?

PR #1588, opened 2026-02-02
Add loading of a single level of includes from configuration files

### LINKED ISSUE #1492: Allow loading WebPush secrets from private file
:bulb: **Idea**
Similar to #590, it would be nice to have to ability to load WebPush secrects from file (preferable the file generated by `ntfy webpush keys`). When using NTFY on NixOS there is otherwise no way to *not* make these keys world-readable.

ntfy server

ntninja:
So far I was not able to compile it locally.

Fixes #1492

### DIFF
--- a/cmd/config_loader.go
+++ b/cmd/config_loader.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"errors"
 	"fmt"
 	"github.com/urfave/cli/v2"
 	"github.com/urfave/cli/v2/altsrc"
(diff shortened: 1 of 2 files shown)


ntninja:
Well, suggesting `go install github.com/goreleaser/goreleaser@latest` is a bad idea given it requires over 4GiB to compile and takes who-knows-how-long to build.

ntninja:
Locally built and tested now.

Go remains an appalling minefield of inconsistencies as always, though…

binwiederhier:
It's not Go. It's the CLI library. It's neat but that yaml stuff is wild and confusing

ntninja:
Well, I didn’t interact with the CLI library at all for this, so I’ll take your word on the joy that may have been. For this code it was things like “oh look, *structs* are passed by value, *slices* are semantically passed by value but internally passed by reference, but *maps* are semantically and internally passed by reference” (I fail to see the logic behind this). Or the whole `x := expr` vs `x, ok := expr` thing… Or Go’s focus on concurrency, while making race-conditions the default (unless you can use a channel, another tuckered-on odd-ball feature). I just wished Go had been developed with the consistency of D, rather than the exceptionality-ness of JavaScript. It’d be a great language then, oh well…

What’s your take on the code?

binwiederhier:
I am not interested in adding include support for the server.yml file at this time. Thanks for your contribution.

PR #1595, opened 2026-02-06
fix: add payments dir to docker build

epifeny:
Include the `payments` folder in `Dockerfile-build` so the Docker build can see the `heckel.io/ntfy/v2/payments` package.

Error before this change:

```
cmd/serve.go:21:2: no required module provides package heckel.io/ntfy/v2/payments; to add it:
    go get heckel.io/ntfy/v2/payments
make: *** [Makefile:190: cli-linux-server] Error 1
```


epifeny:
Thanks for the merge!♥

Issue #338, opened 2022-06-21
Some publish requests on ntfy.sh take up to 15 seconds


binwiederhier:
So I've been noticing that every now and then some requests against ntfy.sh had been taking 11-15s (as opposed to <1s). At first I thought it was a problem with the Linux kernel tuning variables (somaxconn, nofile, ...). Then I thought it was nginx. After randomly poking around I found that the `updateStatsAndPrune()` code is likely to blame, because it locks the server mutex for a very long time (or so it appears).

Here's what I saw:

```
curl -sd hi ntfy.sh/mytopic123 > /dev/null  0.00s user 0.00s system 12% cpu 0.055 total
Tue Jun 21 15:04:57 EDT 2022
curl -sd hi ntfy.sh/mytopic123 > /dev/null  0.01s user 0.01s system 23% cpu 0.062 total
Tue Jun 21 15:05:02 EDT 2022
curl -sd hi ntfy.sh/mytopic123 > /dev/null  0.01s user 0.01s system 0% cpu 11.509 total
                                                                           ^^^^^^
```

This happened even when doing it against localhost:11080 (= not through nginx), meaning DNS and nginx could be ruled out.

I briefly turned on trace logging in ntfy and saw this:

```
Jun 21 18:51:05 ntfy.sh ntfy[4169942]: DEBUG Manager: Pruning messages older than 2022-06-21 06:51:05
Jun 21 18:51:20 ntfy.sh ntfy[4169942]: INFO Stats: 525254 messages published, 60415 in cache, 1478 topic(s) active, 1267 subscriber(s), 6045 visitor(s), 1733 mails received (4 successful, 1729 failed), 12 mails sent (12 successful, 0 failed)
```

This corresponds to this block of code: https://github.com/binwiederhier/ntfy/blob/4e29216b5f7fc170f82fe78896dad41a16916eb4/server/server.go#L1114-L1150, which locks the global server mutex here: https://github.com/binwiederhier/ntfy/blob/4e29216b5f7fc170f82fe78896dad41a16916eb4/server/server.go#L1083

Note the timestamps, 18:51:05 + 18:51:20 -- That's 15 seconds to run this code, meaning that all POST/PUT requests have to wait on the lock this entire time. 

This is likely relatively easy to fix, and looking at the code it is obviously pretty inefficient.

binwiederhier:
[...]

PR #1598, opened 2026-02-08
Fix panic in handleSubscribeHTTP when client disconnects during publish

binwiederhier:
(This PR is AI generated, but a human [me] has verified that it is accurate)

## Summary

- **Fixes a server crash** (nil pointer dereference / SIGSEGV) in `handleSubscribeHTTP` caused by a race between a `topic.Publish` goroutine flushing the HTTP response writer and the handler returning after client disconnect.
- Replaces the existing `wlock.TryLock()` hack with a proper `wlock.Lock()` + `closed` flag pattern that correctly waits for in-flight writes to finish and prevents future writes to a cleaned-up response writer.

## Root cause

`topic.Publish()` copies the subscribers map and calls each subscriber in its own goroutine. When a client disconnects:

1. `handleSubscribeHTTP` returns, triggering deferred cleanup
2. Go's HTTP server cleans up the response writer
3. But a `Publish` goroutine may still be in-flight, calling `Flush()` on the now-invalid writer

The old `TryLock()` approach had two failure modes:
- **TryLock succeeds** (no `sub` running): future `sub()` calls deadlock on `wlock.Lock()` (goroutine leak, since the lock is never released)
- **TryLock fails** (`sub` is running): the handler returns immediately without waiting, and Go cleans up the writer while `sub` is still flushing — exactly the panic

## Fix

The new approach:
- **Deferred cleanup** uses a blocking `wlock.Lock()` to wait for any in-flight `sub()` call to finish, then sets `closed = true`
- **`sub()` function** checks the `closed` flag after acquiring the lock — if the connection has been closed, it returns without touching the response writer

This guarantees the response writer is never accessed after the handler returns.

## Stack trace this fixes

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x5b3525]

goroutine 4355370 [running]:
bufio.(*Writer).Flush(0xc061282940)
net/http.(*response).FlushError(0xc0329b7ce0)
net/http.(*response).Flush(0xc04935ba10?)
heckel.io/ntfy/v2/server.(*Server).handleSubscribeHTTP.func2(...)
        server/server.go:1454
heckel.io/ntfy/v2/server.(*topic).Publish.func1.1(...)
        server/topic.go:117
```

## Test plan

- [x] `go vet ./server/` passes
- [ ] Verify with `make check`
- [ ] Soak test under load with concurrent publishers and disconnecting subscribers

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -1458,12 +1458,15 @@ func (s *Server) handleSubscribeHTTP(w http.ResponseWriter, r *http.Request, v *
 		return err
 	}
 	var wlock sync.Mutex
+	var closed bool
 	defer func() {
-		// Hack: This is the fix for a horrible data race that I have not been able to figure out in quite some time.
-		// It appears to be happening when the Go HTTP code reads from the socket when closing the request (i.e. AFTER
-		// this function returns), and causes a data race with the ResponseWriter. Locking wlock here silences the
-		// data race detector. See https://github.com/binwiederhier/ntfy/issues/338#issuecomment-1163425889.
-		wlock.TryLock()
+		// This blocks until any in-flight sub() call finishes writing/flushing the response writer,
+		// then marks the connection as closed so future sub() calls are no-ops. This prevents a panic
+		// from writing to a response writer that has been cleaned up after the handler returns.
+		// See https://github.com/binwiederhier/ntfy/issues/338#issuecomment-1163425889.
+		wlock.Lock()
+		closed = true
+		wlock.Unlock()
 	}()
 	sub := func(v *visitor, msg *message) error {
 		if !filters.Pass(msg) {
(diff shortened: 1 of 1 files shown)


binwiederhier:
This is strongly related to this issue from 3 years ago: https://github.com/binwiederhier/ntfy/issues/338#issuecomment-1163425889 

I am verifying that this fix is accurate and correct.

binwiederhier:
I verified by THINKING that this is a good fix, and it actually does likely fix a goroutine leak as well. 

Cursor wrote a fantastic test `TestServer_SubscribeHTTP_NoWriteAfterHandlerReturn` to test this as best as possible. I struggled back then to reproduce this because it only happens in a very narrow time window. This test is much better because it reproduces the "write after close" situation reliably, albeit not exactly like before. 

I verified that the test fails without the fix.

PR #1619, opened 2026-02-20
Add experimental PostgreSQL support as alternative database backend

### LINKED ISSUE #1114: Postgres database option
:bulb: **Postgres!**
I use ntfy in a kubernetes environment and would like the option of high availability. I also prefer to use a central DB connection over a volume with a sqlite DB.

:computer: **ntfy server**

binwiederhier:
## Summary

Add PostgreSQL as an alternative database backend to SQLite for all three database-backed stores: message cache, user/access control, and web push subscriptions. Configured via a single `database-url` option (connection string).

Resolves #1114.

## What changed

**New `database-url` config option** — when set to a PostgreSQL connection string (e.g. `postgres://user:pass@host:5432/ntfy`), ntfy uses PostgreSQL for everything instead of SQLite. The `cache-file`, `auth-file`, and `web-push-file` options must not be set when using PostgreSQL. Setting `database-url` implicitly enables auth (equivalent to setting `auth-file`).

### Architecture

- **Shared connection pool**: A single `*sql.DB` PostgreSQL connection pool is created at startup and shared across all stores, rather than each store managing its own connection
- **New packages** extracted from server internals:
  - `model/` — shared types (`Message`, `Attachment`, `Action`, etc.) used across packages
  - `message/` — message cache with SQLite and PostgreSQL backends
  - `webpush/` — web push subscription store with SQLite and PostgreSQL backends  
  - `db/` — shared database utilities (`ExecTx`, `QueryTx`) and PostgreSQL connection helper (`db/pg`)
  - `user/` — refactored manager with separate SQLite and PostgreSQL backends (previously had SQLite-only migrations baked in)
- **Each store** has a `_sqlite.go` / `_postgres.go` split for backend-specific SQL and schema, with shared logic in the main file
- **Schema management**: PostgreSQL schemas are applied on startup (no incremental migrations yet — this is v1/experimental)

### Tests

- All server tests (`server/server_test.go`, `server/server_*_test.go`) run against **both SQLite and PostgreSQL** via `forEachBackend()` test helper
- Individual store tests (`message/cache_test.go`, `user/manager_test.go`, `webpush/store_test.go`) also run against both backends
- CI pipeline (`test.yaml`) spins up a PostgreSQL 17 service container and sets `NTFY_TEST_DATABASE_URL`

### Migration tool

- `tools/pgimport/` — one-off CLI tool to import existing SQLite data (cache, users, web push) into PostgreSQL
- Validates schema versions before import, verifies row counts after
- Idempotent and safe to re-run

### Docs

- New [Database options](https://docs.ntfy.sh/config/#database-options) section documenting SQLite vs PostgreSQL setup
- Connection pool tuning parameters (`pool_max_conns`, `pool_max_idle_conns`, `pool_conn_max_lifetime`, `pool_conn_max_idle_time`)
- Updated message cache, access control, and web push sections to reference PostgreSQL alternative

## Config example

```yaml
base-url: "https://ntfy.example.com"
listen-http: ":2586"
database-url: "postgres://ntfy:mypassword@db.example.com:5432/ntfy?sslmode=require"
auth-default-access: "deny-all"
```

## Test plan

- [x] All existing server tests pass against both SQLite and PostgreSQL
- [x] Store-level unit tests pass for both backends
- [x] CI runs tests with PostgreSQL 17 service container
- [x] Manual testing with `pgimport` migration tool
- [x] Manual testing of fresh PostgreSQL setup end-to-end

### DIFF
--- a/docs/releases.md
+++ b/docs/releases.md
@@ -6,11 +6,47 @@ and the [ntfy Android app](https://github.com/binwiederhier/ntfy-android/release
 
 | Component        | Version | Release date |
 |------------------|---------|--------------|
-| ntfy server      | v2.17.0 | Feb 8, 2026  |
-| ntfy Android app | v1.23.0 | Deb 22, 2026 |
+| ntfy server      | v2.18.0 | Mar 7, 2026  |
+| ntfy Android app | v1.24.0 | Mar 5, 2026  |
 | ntfy iOS app     | v1.3    | Nov 26, 2023 |
 
 Please check out the release notes for [upcoming releases](#not-released-yet) below.
+
+### ntfy server v2.18.0
+Released March 7, 2026
+
+This is the biggest release I've ever done on the server. It's 14,997 added lines of code, and 10,202 lines removed, all from
+one [pull request](https://github.com/binwiederhier/ntfy/pull/1619) that adds [PostgreSQL support](config.md#postgresql-experimental).
+
+The code was written by Cursor and Claude, but reviewed and heavily tested over 2-3 weeks by me. I created comparison documents,
+went through all queries multiple times and reviewed the logic over and over again. I also did load tests and manual regression tests,
+which took lots of evenings.
+
+I'll not instantly switch ntfy.sh over. Instead, I'm kindly asking the community to test the Postgres support and report back to me
+if things are working (or not working). There is a one-off migration tool (entirely written by AI) that you can use to migrate.
+
+**Features:**
+
+* Add experimental [PostgreSQL support](config.md#postgresql-experimental) as an alternative database backend (message cache, user manager, web push subscriptions) via `database-url` config option ([#1114](https://github.com/binwiederhier/ntfy/issues/1114)/[#1619](https://github.com/binwiederhier/ntfy/pull/1619), thanks to [@brettinternet](https://github.com/brettinternet) for reporting)
+
+**Bug fixes + maintenance:**
+
+* Preserve `<br>` line breaks in HTML-only emails received via SMTP ([#690](https://github.com/binwiederhier/ntfy/issues/690), [#1620](https://github.com/binwiederhier/ntfy/pull/1620), thanks to [@uzkikh](https://github.com/uzkikh) for the fix and to [@teastrainer](https://github.com/teastrainer) for reporting)
+
+### ntfy Android v1.24.0
+Released March 5, 2026
+
+This is a tiny release that will revert the "reconnecting ..." behavior of the foreground notification. Lots of people
+have complained about it, so I'm replacing it with a notification that shows up when the server connection has failed
+for >15 minutes, hoping that people will be less annoyed by that.
+
+**Features:**
+
+* Show notification when connection to server has been lost for 15+ minutes, with dismiss, snooze and never-show-again actions
+
+**Bug fixes + maintenance:**
+
+* Fix crash in settings when fragment is detached during backup/restore or log operations
 
 ## ntfy Android v1.23.0
 Released February 22, 2026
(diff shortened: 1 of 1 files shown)


message/store_sqlite_schema.go
```diff
@@ -0,0 +1,466 @@
+package message
```
binwiederhier: Verified that this schema is identical

webpush/store_sqlite.go
```diff
@@ -0,0 +1,142 @@
+package webpush
```
binwiederhier: Manually compared with main sqlite impl and postgres impl

message/store_postgres_schema.go
```diff
@@ -0,0 +1,88 @@
...
+import (
+	"database/sql"
+	"fmt"
+)
+
+// Initial PostgreSQL schema
+const (
+	postgresCreateTablesQuery = `
+		CREATE TABLE IF NOT EXISTS message (
+			id BIGSERIAL PRIMARY KEY,
```
ikopke23: BIGSERIAL over generate_uuid? 
You have timestamps so the order is not super relevant? Is this a backwards compatibility thing since the previous instance used an `Auto Incrementing` type?

server/server.go
```diff
@@ -172,21 +178,38 @@ func New(conf *Config) (*Server, error) {
...
 		if err != nil {
 			return nil, err
 		}
 	}
-	topics, err := messageCache.Topics()
+	topicIDs, err := messageCache.Topics()
 	if err != nil {
 		return nil, err
 	}
+	topics := make(map[string]*topic, len(topicIDs))
```
ikopke23: Until.map no?

db/db.go
```diff
@@ -0,0 +1,38 @@
...
+
+// QueryTx executes a function within a database transaction and returns the result. If the function
+// returns an error, the transaction is rolled back. Otherwise, the transaction is committed.
+func QueryTx[T any](db *sql.DB, f func(tx *sql.Tx) (T, error)) (T, error) {
+	tx, err := db.Begin()
+	if err != nil {
+		var zero T
+		return zero, err
+	}
+	defer tx.Rollback()
```
huronbikes: This will result in closing a transaction twice a la `ExecTx`

db/db.go
```diff
@@ -0,0 +1,38 @@
...
+	if err := f(tx); err != nil {
+		return err
+	}
+	return tx.Commit()
+}
+
+// QueryTx executes a function within a database transaction and returns the result. If the function
+// returns an error, the transaction is rolled back. Otherwise, the transaction is committed.
+func QueryTx[T any](db *sql.DB, f func(tx *sql.Tx) (T, error)) (T, error) {
```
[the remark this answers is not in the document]
binwiederhier: One returns T and the other one doesn't. They are very similar but not the same.

db/db.go
```diff
@@ -0,0 +1,38 @@
...
+)
+
+// ExecTx executes a function within a database transaction. If the function returns an error,
+// the transaction is rolled back. Otherwise, the transaction is committed.
+func ExecTx(db *sql.DB, f func(tx *sql.Tx) error) error {
+	tx, err := db.Begin()
+	if err != nil {
+		return err
+	}
+	defer tx.Rollback()
```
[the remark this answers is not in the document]
huronbikes: An atomic bool in what code? Structuring the execution of code at this abstraction level around the details of a specific implementation seems a bit dangerous.

`Commit()` failing should already abort the transaction in progress, though if a specific implementation of `Commit` errors after the database successfully commits, that's a) going to be a bug of the specific implementation and b) not going to be possible to rollback anyways.  Always invoking `Rollback()` in this functionality will introduce a situation where "it works until it doesn't", eg if a driver implementation does not work in the way this code expects it to.

Issue #690, opened 2023-03-30
html-only emails allow publishing


teastrainer:
Emails can be used to publish messages via ntfy. But html-only mails are rejected with the following error message "554 5.0.0 Error: transaction failed, blame it on the weather: unsupported content type (in reply to end of DATA command)"

That's obviously for security reasons, which are understandable (potentially active / malicious code hidden in the text).

IIRC, emails consisting of text and html text are processed by ignoring the html part.

But sometimes we cannot change the structure of an email, especially if we want to use emails from home servers, home automation etc. Fritz Box for example can be used for email alerts, but sends html-only mails.

I think, there could be two ways to solve this problem:

a) fall back solution (as for emails consisting of text only and html-text):
Ignore or delete html part, meaning: delete / forget the body text completely. Then only the subject would be left for additional information. This would be completely acceptable to me. The text could be replaced by a warning that it has been removed (to avoid too many questions as to why this was done).

b) remove html tags with the help of regex.
This could be done by a (multiline-) search for "(?s)<.*?>" and replacing it with nothing.

That would be an operation with a sledgehammer, as the text is not preserved completely, especially references / links are removed, too. But this is quite acceptable for me. I'll attach the source code of an email from my fritz box before and after processing (in geany text editor)

[1 email source text for test with ntfy.txt](https://github.com/binwiederhier/ntfy/files/11116977/1.email.source.text.for.test.with.ntfy.txt)
[2 email source text for test with ntfy after processing with regex.txt](https://github.com/binwiederhier/ntfy/files/11116978/2.email.source.text.for.test.with.ntfy.after.processing.with.regex.txt)

binwiederhier:
I researched and dabbled with it a bit, and I even had an (infuriatingly bad) conversation with ChatGPT about it (ha! the new world!), and I have decided that there is no safe and easy way to strip HTML tags using regex or other simple means. ChatGPT gave me a few examples showing how stripping with regex could be dangerous. 

Anyway, I looked at [bluemonday](https://github.com/microcosm-cc/bluemonday), and it seems that it doesn't pull in a giant chain of other dependencies, so I think it'll be fine to use it for HTML tag stripping. 

I'd be happy to accept PRs and/or may do it myself some day when I am bored. 

One important note about potential PRs: I do think we should prefer text/plain emails over text/html+stripping, which will change the parsing logic a little.

teastrainer:
[...]

PR #1620, opened 2026-02-21
fix(smtp): preserve <br> line breaks in HTML emails

uzkikh:
## Context
HTML-only emails (e.g. from Synology DSM 7.3 Task Scheduler) use `<br>` tags for line breaks. The existing implementation passed the raw HTML body to bluemonday with `AddSpaceWhenStrippingTag(true)`, which replaces every tag including `<br>` with a space. This caused all notification content to appear on a single unreadable line.

## The ntfy notification before the fix
Task Scheduler has completed a scheduled task.  Task: daily-backup  Start time: Mon, 01 Jan 2026 02:00:00 +0000  Current status: 0 (Normal)  From MyNAS

## The ntfy notification after the fix
Task Scheduler has completed a scheduled task.

Task: daily-backup
Start time: Mon, 01 Jan 2026 02:00:00 +0000
Current status: 0 (Normal)

From MyNAS

## Steps to reproduce
```bash
# Start ntfy with SMTP enabled
ntfy serve --smtp-server-listen :25 --smtp-server-domain example.com

# Send an HTML-only email with <br> line breaks
swaks \
  --server 127.0.0.1 --port 25 \
  --from sender@example.com \
  --to ntfy-mytopic@example.com \
  --header "Content-Type: text/html; charset=utf-8" \
  --body "Line one.<BR>Line two.<BR><BR>Line three."

# Check the notification — all text appears on one line:
#    "Line one. Line two.  Line three."
```

## Root cause

The `readHTMLMailBody()` function used `bluemonday.StrictPolicy().AddSpaceWhenStrippingTag(true)` which replaces  all tags including `<br>` with spaces, discarding line break semantics.

## Fix

Convert `<br>` tags to newlines before passing to bluemonday. The regex covers all variants: `<br>`, `<BR>`, `<br/>`, `<BR />`.

## Tests

- Updated `TestSmtpBackend_HTMLOnly_FromDiskStation` and `TestSmtpBackend_HTMLEmail` to reflect correct output
- Added `TestSmtpBackend_HTMLEmail_BrTagsPreserved` covering the Synology Task Scheduler email format

Related to #690

### DIFF
--- a/server/smtp_server.go
+++ b/server/smtp_server.go
@@ -33,6 +33,7 @@ var (
 var (
 	onlySpacesRegex          = regexp.MustCompile(`(?m)^\s+$`)
 	consecutiveNewLinesRegex = regexp.MustCompile(`\n{3,}`)
+	htmlLineBreakRegex       = regexp.MustCompile(`(?i)<br\s*/?>`)
 )
 
 const (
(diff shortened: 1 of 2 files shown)


binwiederhier:
This is a reasonable fix. Thank you @uzkikh and Cursor/Claude :-)

PR #1623, opened 2026-02-24
feat(web): add configurable datetime format preference

### LINKED ISSUE #1507: Request | Add option to change DateTime format in Web & Mobile apps
:bulb: **Idea**
Add an option to change the datetime format in the web and mobile apps.

Makes sense to have the default match locale, but having an option to manually choose would be awesome. For example I'm from Australia, but I much prefer to use ISO8601 date format (`YYYY-MM-DD`).

- Ntfy Web App
- Android, iOS

scarf005:
## Summary

Resolves #1507

## Testing

image

related: https://github.com/binwiederhier/ntfy-android/pull/162


binwiederhier:
It is not clear to me why this is necessary. Chrome and Firefox on Android and on desktop do not have a setting like this? Do they display time correctly in your locale? I don't really want this to be a thing. It seems that this should be automatically use the system's data format locale?

scarf005:
> Chrome and Firefox on Android and on desktop do not have a setting like this?
> It seems that this should be automatically use the system's data format locale?

It's not possible to separate the language locale (e.g., en-US) from the [date format](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat) in browsers.

> Do they display time correctly in your locale?

I think it's more about personal preference than correctness. user in #1507 said they live in Australia (defaults to `DD/MM/YYYY`) but prefers `YYYY-MM-DD`.

> I don't really want this to be a thing. 

Could you elaborate why? #1507 has been left open and seemed like a valid request.

Newtomic:
@binwiederhier
[comment](https://github.com/binwiederhier/ntfy/pull/1623#issuecomment-3956390728)

---
> It is not clear to me why this is necessary.

Firstly because the **ntfy** isn't using the system format 100% correctly, but also because users (like me) and @scarf005 and others apparently so... too, and ultimately because... please, lol. :-D

---
> Chrome and Firefox on Android and on desktop do not have a setting like this?

Browser's are weird apps, anyway... in my PCs browsers **ntfy**, isn't respecting the language settings nor the system (Windows) settings it's simply always showing in US format, dates like (MM/DD/YY) and times are 12H with AM/PM, and frankly it's not good, on android I like to use English US (because the translations aren't always very good) but even after altering to my language the date format in **ntfy** still uses (DD/MM/YY) format instead of the system configured format (DD/MM/YYYY).

---
> Do they display time correctly in your locale?

Like I said before, not always, besides I want to be able to set my device in one language and still have the option to use a different time and date formats, for example, I do that on my Android and PCs devices like I said before.

---
> I don't really want this to be a thing.

Like @scarf005 asked/said can you say why? Please?

---
> It seems that this should be automatically use the system's data format locale?

Yes sure for the default option I agree with you, but please please provide options to change that globally in the server and/or in the web GUI and android app, I use your app "[android ntfy](https://github.com/binwiederhier/ntfy-android)".

---
Anyway, ultimately it's your call, but please consider these opinions/requests.

Thanks for the **ntfy** project it's awesome!

binwiederhier:
This was done elsewhere

PR #1629, opened 2026-02-28
docs: fix references in access token examples

azrikahar:
- The access token section mentions a `backup-service` user, but it was referred as `ben` in the env variables example
- The private instance section mentions a `backup-script` user, but it's corresponding auth-access is set as `backup-service`
- Additionally mention the token that is included in the private instance section example, just like the access token section


binwiederhier:
Thanks!

PR #1630, opened 2026-03-01
Add debounce-mechanism to messages

### LINKED ISSUE #1550: Debounce messages
:bulb: **Idea**
Add a debounce mechanism to prevent flooding messages.
Introduce two new parameters:
- `debounce_key`: this will be used to group messages that should be debounced
- `debounce_interval`: a period (like the current `delay` parameter)

Every message with the same `debounce_key` will be dropped if another message was successfully sent (not drropped) in the `debounce_interval` interval.

ntfy server

BeeJay28:
Fixes #1550

Default-config-value: time.Duration(0), implying no debounce is used


binwiederhier:
Thank you for your PR. However, I'd rather do a web app debounce, where I don't play a sound when the last message is <1s ago.

Also, your map is never cleared, so it'll eventually OOM most likely

BeeJay28:
> Thank you for your PR. However, I'd rather do a web app debounce, where I don't play a sound when the last message is <1s ago.
> 

Oh, so you mean on client-side in general just silently accept the message as subscriber and send no notification alert? Frankly, I am not too sure what the author of the original issue exactly needs but I guess that'll do too.

> Also, your map is never cleared, so it'll eventually OOM most likely

Oops, that is a very good point. I could just check it on every request or hook it into some goroutine, kind of like you print a status-log on the server-instance every X seconds. All of that only if I would still be going for my original server-side way that is.

binwiederhier:
I implemented the client side "debounce" here: https://github.com/binwiederhier/ntfy/commit/109271a93049e544d8b2e5c9cf4e6098700fa11c

PR #1633, opened 2026-03-02
add --cert for cli

Tyler-Hardin:
I use mTLS for my self hosted server. Would appreciate if you would include support in mainline. Let me know if you see any changes you'd like. Included changes:

* `--cert` flag
* env var `NTFY_CERT`
* client config keys `cert-file` and `cert-password`

I went with `software.sslmate.com/src/go-pkcs12` for pkcs12/cert parsing because the std lib doesn't support sha2, which is standard now.

Thanks for your work on this. Very neat project!

Full disclosure, I used Claude for this. I'm a dev myself, but little experience in Go. I mostly work in Python and Rust. But I reviewed the changes and they do seem sensible to me.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -4,6 +4,7 @@ package client
 import (
 	"bufio"
 	"context"
+	"crypto/tls"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -11,10 +12,13 @@ import (
 	"heckel.io/ntfy/v2/util"
 	"io"
 	"net/http"
+	"os"
 	"regexp"
 	"strings"
 	"sync"
 	"time"
+
+	"software.sslmate.com/src/go-pkcs12"
 )
 
 const (
@@ -34,6 +38,7 @@ var (
 type Client struct {
 	Messages      chan *Message
 	config        *Config
+	http          *http.Client
 	subscriptions map[string]*subscription
 	mu            sync.Mutex
 }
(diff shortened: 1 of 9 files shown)


binwiederhier:
Thanks for the PR. Happy to support mTLS, but I don't want to add another library for that. I highly doubt that the stdlib is not supporting sha256+512 and such.

Tyler-Hardin:
https://github.com/golang/go/issues/22163

Seems like they stopped developing the main lib almost a decade ago.

Grok confirms: https://grok.com/share/bGVnYWN5LWNvcHk_08dce699-f933-483f-83f3-2d2f230d7408

Tyler-Hardin:
From the lib I used:

> This package is forked from golang.org/x/crypto/pkcs12, which is frozen

https://github.com/SSLMate/go-pkcs12

binwiederhier:
Don't believe Grok ;-) Friends don't let friends use Grok.

Here's a version from ChatGPT using the stdlib, prompted on my phone:
https://chatgpt.com/share/69a62512-2c84-800d-be5a-75516a4eee58

I have personally used mTLS in Go using just the stdlib. SHA-256 + SHA-512 are standards. Of course Go supports that.

Tyler-Hardin:
Isn't that a bit of a downgrade? p12 is the standard for mTLS because it encapsulates everything in one file and can be password protected. That's 3 flags/config lines and no password protection possible (PEM is plaintext).

Ask gpt what the standard way to import a p12 is in Go. I quit using it due to the Dept of War deal.

Issue #894, opened 2023-09-24
Re-add tzdata to Dockerfile


binwiederhier:
https://github.com/binwiederhier/ntfy/commit/86c132f9cd4717eaa3c18753ac0f6e44d7c903bc
https://discord.com/channels/874398661709295626/874398661709295629/1152075803102810183

binwiederhier:
Done in https://github.com/binwiederhier/ntfy/commit/8adb9ee633c5cca607892b9ab42b2685fa7fec73

PR #1635, opened 2026-03-03
add tzdata to arm image

abno85:
Small fix: The `tzdata` package was missing from the ARM image so the `TZ` env had no effect after moving my stack to a Raspberry Pi.

Issue #894 only re-added tzdata for the amd64 image.


abno85:
... I must have missed the comment about Alpine not supporting tzdata on ARM, but I built a new image this morning including `tzdata` without any issues and ntfy is using the correct timezone.

PR #1643, opened 2026-03-07
Dockers v2

binwiederhier:
Apparently dockers_v2 is alpha, so I won't switch to it yet. https://goreleaser.com/customization/dockers_v2/#customization


binwiederhier:
Not needed because it's still experimental with goreleaser

PR #1650, opened 2026-03-11
Create Init service file for OpenRC

ageru:
Hi 👋

This in an init file for OpenRC systems.

It should be equivalent in features to the current systemd file, with 2 deliberate changes:
- removed "no-log-dates", as the logs are fine as-is
- Lower nofile limit, as it seems largely sufficient for a self-hosted instance. Feel free to increase to 8192 or 10240 if necessary.

Confirmed functional with Gentoo amd64 and ntfy 2.17.0.

Since I had to create it to get ntfy running on my server, I thought I would spare others the trouble...

Kind regards,

ageru


binwiederhier:
What's the benefit of including this in the main repo? Would we not want to document this in the docs somewhere?

ageru:
The systemd unit file for ntfy server being in the main repo, it seemed a good idea to add this OpenRC init script alongside it, since it serves the same purpose, just on a different init.

I commented the code with installation steps, but I can see it would be more appropriate within the docs. How about enriching the section https://docs.ntfy.sh/install/#linux-binaries ? I could develop a bit that section, adding a few steps:
- create the ntfy user/group
- install the ntfy server service on systemd/OpenRC
- launch the service on systemd/OpenRC

That would be helpful to people that happen to be on an systemd/OpenRC distro without an ntfy package.

binwiederhier:
The reason the systemd file is there is because we build deb+rpm packages that include it. The openrc file would just be entirely lost/dangling, so referencing it in the docs may be a good idea. The listed section sounds good.

server/ntfy-server.openrc
```diff
@@ -0,0 +1,49 @@
+#!/sbin/openrc-run
```
binwiederhier: Should this file be renamed to ntfy.openrc to matc the systemd service file

binwiederhier:
Thank you

PR #1668, opened 2026-03-21
Update web app FAQ to include `require-login`

nihalgonsalves:
This FAQ is outdated, there's the `require-login` option now.


binwiederhier:
thanks

PR #1683, opened 2026-03-30
Restore VirtualHost Example for Apache HTTP Server 2.4.46 and Earlier

jejo86:
In commit 76667ffc the VirtualHost example configuration was adapted to use the latest ProxyPass parameter 'upgrade'. Restore the previous example for Apache HTTP Server 2.4.46 and earlier, which do not have this function yet without replacing the more current solution, but having both side-by-side. 

https://httpd.apache.org/docs/2.4/mod/mod_proxy.html#protoupgrade

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -1385,7 +1385,7 @@ or the root domain:
     }
     ```
 
-=== "Apache2"
+=== "Apache >= 2.4.47"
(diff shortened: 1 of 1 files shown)


binwiederhier:
Thanks

PR #1696, opened 2026-04-12
Include Darwin in the serve Unix build

### LINKED ISSUE #1631: cmd: darwin build fails with undefined maybeRunAsService/sigHandlerConfigReload

Building the server-enabled CLI on macOS fails due to missing Darwin-specific symbol implementations in the `cmd` package.

`cmd/serve.go` unconditionally calls `maybeRunAsService` and `sigHandlerConfigReload`, but neither is available for Darwin builds, causing compilation to fail.

ntfy server (CLI build path), specifically `make cli-darwin-server`

Reproduction command:

```bash
make cli-darwin-server
```

Build output:

```
# heckel.io/ntfy/v2/cmd
cmd/serve.go:509:26: undefined: maybeRunAsService
cmd/serve.go:517:5: undefined: sigHandlerConfigReload
make: *** [cli-darwin-server] Error 1
```

Root cause appears to be build tags:

`cmd/serve.go` calls both functions unconditionally.
`cmd/serve_windows.go` only builds on `windows && !noserver`.
`cmd/serve_unix.go` only builds on `linux || dragonfly || freebsd || netbsd || openbsd` (Darwin missing).

This likely regressed when Windows service support was introduced (commit 6d5cc6ae).

ShipItAndPray:
## Summary
- include macOS in the Unix serve implementation so Darwin builds get `maybeRunAsService` and `sigHandlerConfigReload`
- keep the change minimal by extending the existing non-Windows build tag instead of introducing a separate Darwin stub

Fixes #1631

## Validation
- git diff --check
- attempted local `go build .` validation on darwin after creating temporary placeholder embed dirs for `server/docs` and `server/site`

## Notes
- the current checkout is missing embedded `server/docs` and `server/site` content, so the local build cannot complete end-to-end without temporary placeholders
- the underlying compile issue from #1631 is addressed by making `cmd/serve_unix.go` build on Darwin as well

### DIFF
--- a/cmd/serve_unix.go
+++ b/cmd/serve_unix.go
@@ -1,4 +1,4 @@
-//go:build linux || dragonfly || freebsd || netbsd || openbsd
+//go:build darwin || linux || dragonfly || freebsd || netbsd || openbsd
 
 package cmd
 


binwiederhier:
This is not complete, as you can see by the build error in the pipeline. I'll still merge it and fix it manually. I think I'll just make the `serve` command work on macOS and call it experimental support so people can try it out if they wanted to.

PR #1697, opened 2026-04-12
Read full stdin when publishing

### LINKED ISSUE #1677: 1MB attachment limit on client and broken pipe when exceeding

First, the client seems to limit the size of an attachment to 1MB, at least when using stdin.
I've set `attachment-file-size-limit: 5M` on the server and in my local client.yml but it doesn't change the behavior.

Additionally, and maybe with a higher severity, when piping the output of a process to `ntfy`, it closes the pipe once the limit is reached, killing the process at the start of the chain.

There should be a way to say to ntfy "OK, attach only data to the size limit and discard the rest, but don't exit when data is still flowing" so that the first process would keep going instead of being killed by a `Broken pipe`.

Self-hosted ntfy, version 2.19.2
Local ntfy client, same version.

ShipItAndPray:
Fixes #1677.\n\n previously truncated redirected stdin at 1 MiB before sending the request. That closed the pipe early and could kill upstream producers with a broken pipe once the payload exceeded the hardcoded cap.\n\nThis removes the premature cap so redirected stdin is consumed in full, and adds a regression test that verifies the entire payload reaches the server.

### DIFF
--- a/cmd/publish.go
+++ b/cmd/publish.go
@@ -123,6 +123,8 @@ func execPublish(c *cli.Context) error {
 	if err != nil {
 		return err
 	}
+	// Stream redirected stdin directly instead of buffering it into memory first.
+	useStdinMessage := file == "" && message == "" && isStdinRedirected()
 	var options []client.PublishOption
 	if title != "" {
 		options = append(options, client.WithTitle(title))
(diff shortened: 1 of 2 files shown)


cmd/publish.go
```diff
@@ -268,7 +268,7 @@ func parseTopicMessageCommand(c *cli.Context) (topic string, message string, com
 	}
 	if message == "" && isStdinRedirected() {
 		var data []byte
-		data, err = io.ReadAll(io.LimitReader(c.App.Reader, 1024*1024))
+		data, err = io.ReadAll(c.App.Reader)
```
binwiederhier: 1. This will cause an OOM. 
2. This is not the right way to do this. There is already a `--file` option that support `--file=-`. If you can make it use that mechanism here then sure.

ShipItAndPray: Reworked this to stream redirected stdin directly into `PublishReader` instead of reading it into memory first, so it no longer risks an OOM or closes the pipe early. The PR branch is also cleaned up now and only contains the publish fix. I added a regression test covering >1 MiB stdin payloads and validated it with `go test ./cmd -run 'TestCLI_Publish_Stdin_ReadsFullInput|TestCLI_Publish_Subscribe_Poll$'` (using temporary local embed placeholders and a Darwin serve stub because `main` currently fails to compile on macOS for an unrelated reason).

PR #1703, opened 2026-04-14
Add en_CA locale for Canadian English

usefulish:
The ntfy web UI formats dates using Intl.DateTimeFormat(i18n.language). This means locale subtags (like en-CA) already produce correct regional date formats without any additional logic.

Currently, only en is shipped, so browsers set to en-CA fall back to en, resulting in US-style MM/DD/YYYY dates.

Adding en_CA.json allows users with en-CA browsers (or those selecting it manually) to get the correct YYYY-MM-DD date format automatically.

This change:
	•	does not modify any application logic
	•	does not affect existing English (en) users
	•	simply adds a locale file (currently identical to en.json)


binwiederhier:
You didn't add a locale, you renamed the default en.json to en_CA.json

PR #1720, opened 2026-04-28
Fix opening links with noreferrer

dmitrylyzo:
`rel="noreferrer"` has the same effect as `rel="noopener"`, but also prevents the `Referer` header from being sent to the new page.
https://mui.com/material-ui/react-link/#security

If this feature is set, the browser will omit the `Referer` header, as well as set `noopener` to true.
https://developer.mozilla.org/en-US/docs/Web/API/Window/open#noreferrer

**Changes**
Drop `noopener` if `noreferrer` is set.

**Issues**
N/A
The `Referer` header is sent to the server.

**Notes**
The value of the `rel` attribute must be space-separated.
https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/rel

### DIFF
--- a/web/src/app/utils.js
+++ b/web/src/app/utils.js
@@ -178,7 +178,7 @@ export const formatPrice = (n) => {
 };
 
 export const openUrl = (url) => {
-  window.open(url, "_blank", "noopener,noreferrer");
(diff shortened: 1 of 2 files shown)


binwiederhier:
thanks

PR #1725, opened 2026-05-04
Add X-Persist header to control client-side local storage

EugeneShtoka:
## Summary

Introduces a new `X-Persist` (or `Persist`) publish parameter. When set to `no`, the server includes `"persist": false` in the broadcast JSON payload, signalling clients that they should not save the message to local history.

This is intentionally **separate from `X-Cache`** (server-side caching) so that publishers can control client persistence independently:

- `X-Cache: no` — server does not cache the message
- `X-Persist: no` — clients should not save the message to local history

A common use case is ephemeral clipboard sync or other transient notifications that should be delivered and shown, but not accumulate in message history.

## Changes

- `model/model.go` — add `Persist *bool` field to `Message` (omitted from JSON when nil)
- `server/util.go` — add `readOptionalBoolParam` helper returning `*bool` (nil when not set)
- `server/server.go` — parse `x-persist`/`persist` header and set `m.Persist`

## Related

Android client support: EugeneShtoka/ntfy-android#1 (respects `"persist": false` to skip local DB insert)

### DIFF
--- a/model/model.go
+++ b/model/model.go
@@ -47,6 +47,7 @@ type Message struct {
 	PollID      string      `json:"poll_id,omitempty"`
 	ContentType string      `json:"content_type,omitempty"` // text/plain by default (if empty), or text/markdown
 	Encoding    string      `json:"encoding,omitempty"`     // Empty for raw UTF-8, or "base64" for encoded bytes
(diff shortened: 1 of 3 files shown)


binwiederhier:
Thank you for your contribution, but I don't think I want to add this feature. Please check with me on Discord/Matrix before working on new features.

Issue #1498, opened 2025-11-25
[feature request] heartbeat monitoring on the mobile apps


giorgiga:
:bulb: **Idea**

Have the mobile apps notify the user is a message is *not* received on a specific channel during a specified timeframe.

This would allow all of us using ntfy for system monitoring/alerting the peace of mind of being automatically notified if our monitoring/alerting system can't send notifications.

The "no heartbeat" notification should be suppressed if no network is enabled on the phone (eg. because it's in plane mode).

Ideally, the user should be able to select whether a degraded/faulty connection (eg. mobile data enabled but no mobile network coverage, or being connected to a wifi that is isolated from the Internet) suppresses the "no heartbeat" notification or not.

:computer: **Target components**

mobile apps

PR #1738, opened 2026-05-12
Add server-side heartbeat monitor feature

fluffyspace:
## Summary

Adds an opt-in heartbeat / dead-man's-switch subsystem to the ntfy server. Authenticated users register a monitor with a key, period, grace, and alert topic; the script/service being watched POSTs to `/v1/heartbeat/<key>` on a schedule, and a missed window causes ntfy to publish a DOWN alert on the configured topic (recovery UP alert when pings resume).

This is opt-in: with no `MonitorFile` configured, the routes return `40066 monitors not enabled` and the background goroutine does not start, so behavior is identical to current ntfy for users who don't enable it.

I built and deployed this on my own server and have been pinging from a real cron'd script for a few hours — DOWN and UP alerts both fire reliably end-to-end (Android notification, message in topic, web push, all work).

## Context / prior art

I'm aware of:
- #1498 (open feature request) — but that one is specifically about *mobile-app-side* absence detection, which this PR does not address. The use cases overlap, but the implementation surface is completely different.
- #1425 (merged) — added docs pointing to the external `ntfy-heartbeat-monitor` tool, which polls topics from outside ntfy.

I'm submitting this in case a built-in option is interesting; happy to close if the project's direction is firmly "use external tools." The motivation for me was simplicity (one fewer moving part for self-hosters who already have ntfy) and that the external tool model requires another service that itself needs monitoring.

## API

All endpoints require authentication and are scoped per user:

| Method  | Path                       | Purpose                  |
|---------|----------------------------|--------------------------|
| `POST`  | `/v1/monitors`             | Create monitor           |
| `GET`   | `/v1/monitors`             | List user's monitors     |
| `GET`   | `/v1/monitors/<key>`       | Get one                  |
| `DELETE`| `/v1/monitors/<key>`       | Delete                   |
| `GET`/`POST` | `/v1/heartbeat/<key>` | Record heartbeat         |

Create body: `{key, period, grace, alert_topic, alert_priority?}`

## How alerts are delivered

`publishMonitorAlert` mirrors `sendDelayedMessage` — it looks up the topic, attributes the message to the monitor owner (via the existing `userManager`), and goes through the normal publish path: forward to active subscribers, cache for offline subscribers, Firebase, Web Push.

## State machine

- New monitor: `pending` (no alerts yet — first ping required)
- Heartbeat arrives: `pending|down → up`. On `down → up`, recovery alert published.
- Background ticker every `MonitorCheckInterval` (default 30s): any `up` monitor with `now > last_seen + period + grace` flips to `down` and an alert is published.

## Configuration

New CLI flags / env vars:
- `--monitor-file` / `NTFY_MONITOR_FILE` (empty = feature off)
- `--monitor-startup-queries` / `NTFY_MONITOR_STARTUP_QUERIES`
- `--monitor-check-interval` / `NTFY_MONITOR_CHECK_INTERVAL` (default 30s)

## What's intentionally NOT in this PR

These are deliberate to keep the diff focused; happy to extend if there's interest:

- **No PostgreSQL backend.** The `monitor` package is SQLite-only. Other packages support both via the queries-struct pattern; this could be added later but I wanted to keep the initial diff small.
- **No web UI.** The feature is API-only for now. The web app would benefit from a /monitors page; out of scope here.
- **No user docs.** Will add `docs/config.md` + `docs/publish.md` content if the feature direction is acceptable.
- **No Android-app changes.** This PR is server-side only. The mobile concern in #1498 is separate.

## Test plan

- [x] 27 unit + integration tests added (monitor package + server handlers)
- [x] Full `go test ./monitor/... ./server/...` passes
- [x] Build and `go vet ./...` clean
- [x] Deployed and exercised against a real cron'd script — DOWN and UP alerts both verified end-to-end on Android via ntfy app + Firebase + Web Push
- [x] Verified that with `MonitorFile=""` (default) all endpoints return 40066 and no goroutine starts, so behavior is unchanged for existing deployments

### DIFF
--- a/cmd/serve.go
+++ b/cmd/serve.go
@@ -111,6 +111,9 @@ var flagsServe = append(
 	altsrc.NewStringFlag(&cli.StringFlag{Name: "web-push-startup-queries", Aliases: []string{"web_push_startup_queries"}, EnvVars: []string{"NTFY_WEB_PUSH_STARTUP_QUERIES"}, Usage: "queries run when the web push database is initialized"}),
 	altsrc.NewStringFlag(&cli.StringFlag{Name: "web-push-expiry-duration", Aliases: []string{"web_push_expiry_duration"}, EnvVars: []string{"NTFY_WEB_PUSH_EXPIRY_DURATION"}, Value: util.FormatDuration(server.DefaultWebPushExpiryDuration), Usage: "automatically expire unused subscriptions after this time"}),
 	altsrc.NewStringFlag(&cli.StringFlag{Name: "web-push-expiry-warning-duration", Aliases: []string{"web_push_expiry_warning_duration"}, EnvVars: []string{"NTFY_WEB_PUSH_EXPIRY_WARNING_DURATION"}, Value: util.FormatDuration(server.DefaultWebPushExpiryWarningDuration), Usage: "send web push warning notification after this time before expiring unused subscriptions"}),
+	altsrc.NewStringFlag(&cli.StringFlag{Name: "monitor-file", Aliases: []string{"monitor_file"}, EnvVars: []string{"NTFY_MONITOR_FILE"}, Usage: "file used to store heartbeat monitor state (enables the monitor feature)"}),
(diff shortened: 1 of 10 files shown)


binwiederhier:
ntfy already supports dead man's snitch. Please read the docs. 

Also, before submitting substantial features, please chat with us on discord/matrix. 

Thanks for your interest in ntfy

fluffyspace:
Apologies — you're right, I missed the section in publish.md. Closing this. Thanks for the pointer.

PR #1756, opened 2026-05-27
fix: include exact UTC timestamp in APNs notification subtitle for iOS (#1749)

### LINKED ISSUE #1749: Need exakt time and date when reciving a push
iOS
Need exakt time and date.
Now it just showing yesterday if you check the day after.

rishichintala:
## Summary

- Add the original message time as a formatted UTC subtitle (`SubTitle`) in APNs alert payloads, so iOS always shows the exact date and time in the notification banner — even when the notification center groups older messages under a relative label like "Yesterday"
- Add missing `apns-push-type: alert` and `apns-priority` headers to APNs alert notifications, aligning with Apple's requirements for iOS 13+
- Update Firebase tests to match the new `APNSConfig` structure

## Background

iOS groups notifications from the previous day under "Yesterday" in the notification center, hiding the exact timestamp. Since this is an iOS system UI behavior, it cannot be changed server-side. The fix embeds the exact UTC timestamp directly in the `SubTitle` field of the APNs payload, so the original message time is always visible in the notification content itself.

Closes #1749

## Test plan

- [x] `TestToFirebaseMessage_Message_Normal_Allowed` — verifies SubTitle and headers for normal messages
- [x] `TestToFirebaseMessage_Message_Normal_Not_Allowed` — verifies SubTitle and headers for poll_request fallback
- [x] `TestToFirebaseMessage_PollRequest` — verifies SubTitle and headers for poll_request messages
- [x] All firebase tests pass: `go test ./server/... -run TestToFirebase`


rishichintala:
@binwiederhier Can you look into this?

PR #1768, opened 2026-05-30
security: pin GitHub Actions

nihalgonsalves:
I saw that you added a Dependabot config and cooldown in a3f0f6c.

This adds pinned GitHub Actions to guard against common GitHub Actions security issues. I ran:

```sh
pinact run --min-age=7 --verify-min-age
```

I added the `group` config because it otherwise ends up with a _lot_ of PRs every week even for patch updates.

Dependabot will automatically bump the SHA as well as the `# <tag>` part

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -20,3 +20,7 @@ updates:
       interval: "weekly"
     cooldown:
       default-days: 7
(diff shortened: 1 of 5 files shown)


binwiederhier:
Thank you

Issue #420, opened 2022-09-30
E-mail publishing with access-control enabled


huwylphi:
Hi, I've enabled [access-control](https://ntfy.sh/docs/config/#access-control) and since I host ntfy on a private instance, I set [`auth-default-access` to `deny-all`](https://ntfy.sh/docs/config/#example-private-instance) and added one user with specific permission on a specific topic.
I also enabled the[ e-mail publishing](https://ntfy.sh/docs/config/#e-mail-publishing) feature in order to publish message by sending an e-mail to the ntfy server.
Now when sending an e-mail I get a "forbidden" error feedback. When I switch off the access-control, the e-mails are sent. So I guess this has something to do with authentication and ACL check.
How does the ntfy smtp-server fetch the user/password sending the e-mail? I guess this is needed for checking permission against ACL?
Until now I tried publishing message via e-mail:
1. by passing the user vie the from address (left side of '@'). But if that would be correct, how to pass the password?
2. with smtp authentication

but none these 2 ideas are working.

Any example of e-mail publishing with user authentication or any hints would be appreciated.
Thanks.

binwiederhier:
> How does the ntfy smtp-server fetch the user/password sending the e-mail? I guess this is needed for checking permission against ACL?

Email sending to password-protected topics is currently not supported. I am not entirely sure how to facilitate the login without giving away the password over a potentially insecure channel.

It would have to be part of the e-mail address, otherwise you woudn't be able to hook it up to external tools. Something like (a) `ntfy-mytopic+user+pass@ntfy.sh` or (b) `ntfy-mytopic+token@ntfy.sh` or just (c) `token@ntfy.sh`. 

Option (c) would be the nicest, and would also enable the use of `Bearer` tokens in the HTTP Auth header, but it would mean extra auth work to store and associate tokens. Option (a) would be the easiest to implement, but you'd have your password in the clear and give it to whoever is sending the email.

TLDR: It's not implemented. There's work to implement it, and I'd prioritize other issues before this as of today.

huwylphi:
And what about my idea number 2: using the **SMTP Authentication**? I mean if the ntfy SMTP server would support a well established authorization protocols, then the user and password could be sent when the client is authenticating to the SMTP server?

Otherwise as a workaround, would it be possible to run 2 instances of ntfy that **shares the same cache.db file**, where

- one instance is exposed on the public network for default ACL protected subscribing/publishing and
[...]

PR #1775, opened 2026-06-06
docs: document SMTP AUTH PLAIN for email publishing

buddhaCode:
## What

This documents that ntfy's built-in SMTP server (email publishing) supports **SMTP AUTH PLAIN** and authenticates against the ntfy user database — so email publishing works with `auth-default-access: deny-all` without granting anonymous topic access.

## Why

SMTP AUTH PLAIN was added in #1164 (merged), where the maintainer noted *"I'll add some docs around it."* The config docs were never updated. `publish.md` briefly mentions it (*"you can use SMTP PLAIN auth"*), but `config.md` — where the SMTP server is configured — only documents the token-in-the-address variant for protected topics. This PR closes that gap and explains the *how*. The original question goes back to #420 ("E-mail publishing with access-control enabled").

## Changes

**`docs/config.md`**
- New section **"Authenticating to protected topics"** documenting SMTP AUTH PLAIN, with both auth paths:
  - username + password → validated as a ntfy user
  - empty username + access token as the password → password treated as a token (same as HTTP token auth)
  - small ASCII diagram showing both paths and that a token in the address takes precedence
  - note that the SMTP server does **not** advertise STARTTLS (credentials sent unencrypted)
- Clarify that **MX/A records** are only needed for MX-based delivery from other mail servers, not for clients/appliances that connect directly to the SMTP server's host and port
- New **"Message format and content"** section: subject → title, body → message, plain text only (no Markdown), and how multipart plain/HTML parts are handled

**`docs/publish.md`**
- Link the existing "SMTP AUTH PLAIN" mention to the new config section
- Clarify the subject/body behavior (the previous "only supports adding a message title" wording was misleading)

## Notes

All behavior was verified against `server/smtp_server.go` (`AuthPlain` → `Authorization: Basic`, token-in-address precedence, plain/HTML body handling) and tested against a live instance. Docs-only change; no code changes. Happy to adjust wording.

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -1077,9 +1077,14 @@ Here's an example config (this is how it is configured for `ntfy.sh`):
     smtp-server-addr-prefix: "ntfy-"
     ```
 
-In addition to configuring the ntfy server, you have to create two DNS records (an [MX record](https://en.wikipedia.org/wiki/MX_record) 
-and a corresponding A record), so incoming mail will find its way to your server. Here's an example of how `ntfy.sh` is 
-configured (in [Amazon Route 53](https://aws.amazon.com/route53/)):
+Whether you need DNS records depends on how mail reaches your server. If **other mail servers** deliver to your domain
+via an [MX lookup](https://en.wikipedia.org/wiki/MX_record) (the usual case for receiving regular email), you have to
+create two DNS records: an MX record and a corresponding A record, so incoming mail will find its way to your server.
+If instead your sender connects **directly** to the SMTP server's host and port — e.g. a router's push-mail feature, or
+any appliance where you enter the SMTP server address yourself — no DNS records are needed, and publishing works from
+anywhere the port is reachable. See [local-only email](#local-only-email) below for that case.
+
+The following is an example of how `ntfy.sh` is configured for MX-based delivery (in [Amazon Route 53](https://aws.amazon.com/route53/)):
 
 <figure markdown>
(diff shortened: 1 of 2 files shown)


binwiederhier:
Both code and PR description are clearly AI generated (emdashes, unicode tree, ..). I have no way of telling whether this has been reviewed by you or not. Please review yourself and make everythign sound less like AI.

buddhaCode:
Moin binwiederhier,
You're right. I had the AI generate the doc text and the PR text. I didn't make this transparent because the contribution guidelines for this repo don't mention AI. Claude Code automatically suggests not mentioning it (for fear of rejection ;)). I’m just starting to gain experience with AI contributions. I’ll be more transparent about this in the future, even if it isn’t strictly required.

Nevertheless, the PR’s content stems from real, human “pain”. I simply couldn’t find enough info about SMTP auth in the docs. And I wanted to contribute to the project, because it provides me with real-world benefits. And in the hope that people after me will find more info on the topic.

Linguistically, I didn’t micromanage the AI. In terms of content, I provided guidelines on structure and scope. And I iterated through several rounds until I was satisfied with the length and precision. So I was in the loop.

I’ll revise the language again in the next few days so it doesn’t sound like AI.

Do the changes work for you in terms of content and structure?

Best regards, Arne

buddhaCode:
Reworded by hand: em dashes, the tree diagram and most of the bolding are gone.
I also dropped the "Message format and content" section because it repeated the
paragraph right above it, so the diff is quite a bit smaller now.

PR #1793, opened 2026-06-17
set apns sound for ios notifs

am7590:
## Summary

This updates APNS alert payload generation so visible iOS notifications include an explicit sound.

- sets `aps.sound = "default"` for normal message notifications
- sets `aps.criticalSound` for priority 5 message notifications
- adds tests covering both normal and critical APNS sound behavior

## Testing

```bash
env GOCACHE=/private/tmp/go-build GOMODCACHE=/private/tmp/go-modcache go test ./server -run 'TestToFirebaseMessage_(Message_Normal_Allowed|Message_Critical_Allowed)$'
env GOCACHE=/private/tmp/go-build GOMODCACHE=/private/tmp/go-modcache go test ./server -run 'TestToFirebaseMessage|TestMaybeTruncateFCMMessage'
```

## Todo
- [ ] On-device testing

### DIFF
--- a/server/server_firebase.go
+++ b/server/server_firebase.go
@@ -241,16 +241,28 @@ func createAPNSAlertConfig(m *model.Message, data map[string]string) *messaging.
 	for k, v := range data {
 		apnsData[k] = v
 	}
+	aps := &messaging.Aps{
+		MutableContent: true,
+		Alert: &messaging.ApsAlert{
+			Title: m.Title,
+			Body:  maybeTruncateAPNSBodyMessage(m.Message),
+		},
+	}
+	if m.Event == model.MessageEvent {
+		if m.Priority == 5 {
+			aps.CriticalSound = &messaging.CriticalSound{
+				Critical: true,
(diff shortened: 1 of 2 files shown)


server/server_firebase.go
```diff
@@ -241,16 +241,28 @@ func createAPNSAlertConfig(m *model.Message, data map[string]string) *messaging.
...
+			Body:  maybeTruncateAPNSBodyMessage(m.Message),
+		},
+	}
+	if m.Event == model.MessageEvent {
+		if m.Priority == 5 {
+			aps.CriticalSound = &messaging.CriticalSound{
+				Critical: true,
+				Name:     "default",
+				Volume:   1.0,
+			}
```
binwiederhier, line 257: If the iOS toggle is toggled off, do we still need `aps.Sound = "default"`?. I'll test myself I guess :-D

am7590: It's best practice. May resolve a bug where iOS 26 users have notif sounds break.

Issue #1734, opened 2026-05-11
Ntfy ignores protocol for S3 bucket


Kernald:
:lady_beetle: **Describe the bug**

When parsing the S3 configuration, Ntfy ignores the protocol it's being passed, and hard-codes `https`:
https://github.com/binwiederhier/ntfy/blob/802c0a4c303d76db911d5f83c17320651efdf84d/s3/types.go#L26-L31

When targeting a local S3 provider, this might not be desirable - my Ceph RGW instance doesn't have an internal TLS certificate. Virtually every other service is reaching it over `http` (not `https`) just fine.

:computer: **Components impacted**

Server

:bulb: **Screenshots and/or logs**

:crystal_ball: **Additional context**

binwiederhier:
Done in https://github.com/binwiederhier/ntfy/pull/1794

PR #1794, opened 2026-06-17
Preserve the original S3 endpoint scheme

sskender:
Preserve the original S3 endpoint scheme in Config rather than stripping it. Maintain HTTPS as the default for backward compatibility.

Verified with a local MinIO instance over HTTP — this scenario did not work previously and now behaves correctly.

Refs #1734


binwiederhier:
Thanks. I had to adjust it a tad to make sure it's backwards compatible. But this is great.

PR #1805, opened 2026-06-25
Support multiple attachments per message

TheDuffman85:
## Summary

This PR adds support for multiple attachments per message while keeping the existing single-attachment behavior fully backwards compatible.

Existing clients that only understand the legacy `attachment` field or single-file publish flows continue to work unchanged. New clients can use the canonical `attachments` list for multiple attachments.

## What changed

- Adds canonical `attachments` support to message JSON output
- Keeps legacy `attachment` as a compatibility alias for the first attachment
- Adds multipart publishing with repeated `attachment` file fields
- Adds multipart publishing with repeated `attach` URL fields
- Adds JSON `attachments` support for multiple external URL attachments
- Rejects ambiguous mixed legacy/new attachment input
- Adds a configurable per-message attachment count limit
- Stores attachments in a normalized attachment table
- Migrates existing single attachment cache data
- Updates attachment cleanup, expiry, stats, downloads, and logging
- Adds Firebase payload support for multiple attachments
- Updates docs and config examples

## Tooling support

All publishing tools now support multiple attachments:

- HTTP API via multipart/form-data
- JSON publish API for multiple external URL attachments
- Web UI file picker/drop/paste flows
- Web notification rendering
- ntfy CLI with repeated `--file` and `--attach`
- Go client message types

## Compatibility

This PR is fully backwards compatible:

- Existing raw body upload with `Filename` still works
- Existing `Attach` / `X-Attach` URL attachment publishing still works
- Existing JSON `attach` and `filename` publishing still works
- Existing subscribers can keep using `attachment`
- New subscribers can use `attachments`
- The legacy `attachment` field remains populated from the first attachment

## Mobile follow-up

If this PR is merged, I will also prepare a follow-up PR for the Android mobile app to add multi-attachment publishing support there.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -40,17 +40,18 @@ type Client struct {
 
 // Message is a struct that represents a ntfy message
 type Message struct { // TODO combine with server.message
-	ID         string
-	Event      string
-	Time       int64
-	Topic      string
-	Message    string
-	Title      string
-	Priority   int
-	Tags       []string
-	Click      string
-	Icon       string
-	Attachment *Attachment
+	ID          string
+	Event       string
+	Time        int64
+	Topic       string
+	Message     string
+	Title       string
+	Priority    int
+	Tags        []string
+	Click       string
+	Icon        string
+	Attachment  *Attachment
+	Attachments []*Attachment
 
 	// Additional fields
 	TopicURL       string
--- a/cmd/publish.go
+++ b/cmd/publish.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"bytes"
 	"errors"
 	"fmt"
 	"github.com/urfave/cli/v2"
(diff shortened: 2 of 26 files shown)


binwiederhier:
I don't think I want to support this right now. Feel free to open an issue though and link this PR.

TheDuffman85:
@binwiederhier I can totally understand that. It's a big change. 

**The real problem I'm trying to solve is this:** I use NTFY, among other things, to transfer files between my PC and my phone. But unfortunately, I have to share or send each file individually with NTFY because a notification can only have one attachment.

How about if we added a feature to the app that lets users share multiple files with the app, which would then generate multiple notifications? That would be a small, very localized change. 

By the way, I’m the developer of the “Send to NTFY” browser extension :)

binwiederhier:
Thanks for the "Send to NTFY" extension! 

Don't get me wrong, I like the feature^, but it requires work in Android, iOS and thorough testing. And there are so many more important features. I don't want to spend time on this. 

Create a ticket and get people to :+1: it.

PR #1827, opened 2026-07-08
Return ntfy error code and message in headers on failed WebSocket subscriptions

### LINKED ISSUE #535: Websocket failures are indistinguishable
When you're unauthroised on a websocket, wrong credentials or none provided, the response code is a 200 but the debug log on the server reports unauthorized. I think it should be 403 or 401, not 200.

Additionally, If a proxy fails and returns an incorrect page with a 200, my app says credential failure, when that definitely isn't, it's a connection failure, but it is currently not possible to differentiate this with ntfy.sh's behaviour.

TowyTowy:
Failed WebSocket subscriptions already return the proper HTTP status code (e.g. 401/403) when the failure happens before the upgrade completes — but clients still cannot read *why* it failed (the JSON error body is not consumable by most WebSocket clients, which reject anything that is not a `101`), and they cannot tell a genuine ntfy error apart from a `200` page returned by an intermediary proxy. This is the "indistinguishable" part of #535.

This completes the suggestion from the issue thread by adding `X-Ntfy-Error-Code` (ntfy error code) and `X-Ntfy-Error` (message) response headers in the pre-upgrade WebSocket error path, mirroring the JSON body fields. Their presence lets clients distinguish real ntfy failures from proxy responses. Post-upgrade (hijacked) errors are left untouched via the existing `errWebSocketPostUpgrade` guard, since the connection can no longer carry HTTP headers.

Adds an end-to-end test (auth-failed `/ws` returns 403 + headers) and a unit test covering the pre- vs post-upgrade header behavior; existing hijacked-connection test still green.

Happy to adjust header naming/shape if you prefer a different form.

Fixes #535

### DIFF
--- a/docs/releases.md
+++ b/docs/releases.md
@@ -2078,6 +2078,12 @@ and the [ntfy Android app](https://github.com/binwiederhier/ntfy-android/release
 
 ## Not released yet
 
+### ntfy server v2.28.0 (UNRELEASED)
+
+**Bug fixes + maintenance:**
+
+* Failed WebSocket subscriptions (e.g. auth failures) now return the ntfy error code and message in the `X-Ntfy-Error-Code` and `X-Ntfy-Error` response headers, alongside the proper HTTP status code, so clients can distinguish real failures from a `200` page returned by a proxy ([#535](https://github.com/binwiederhier/ntfy/issues/535), thanks to [@alexhorner](https://github.com/alexhorner) for reporting)
+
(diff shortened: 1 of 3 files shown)


TowyTowy:
Rebased on main to clear the conflict. v2.27.0 shipped without this, so I moved the changelog line into a new `v2.28.0 (UNRELEASED)` block - rename it if that's not the next version you had in mind.

Issue #1803, opened 2026-06-25
2.25.0 `make web` fails in COPR chroots


cyqsimon:
:lady_beetle: **Describe the bug**

I attempted to build 2.25.0 [on COPR](https://copr.fedorainfracloud.org/coprs/cyqsimon/ntfysh/build/10649601/). All chroots are failing at `make web` step.

:computer: **Components impacted**

COPR linux package

:bulb: **Screenshots and/or logs**

```
PWA v1.3.0
Building public/sw.js service worker ("es" format)...
vite v8.1.0 building client environment for production...
make: *** [Makefile:145: web-build] Error 139
error: Bad exit status from /var/tmp/rpm-tmp.dkCW45 (%build)
    Bad exit status from /var/tmp/rpm-tmp.dkCW45 (%build)
```

Full logs: [copr-logs.tar.gz](https://github.com/user-attachments/files/29341289/copr-logs.tar.gz)

:crystal_ball: **Additional context**

- I have not modified the build script in any way since 2.24.0.
- On my ArchLinux box with node-26.3.1 `make web` works just fine. But on all my COPR chroots, the `nodejs` package selects node-22.22.2 by default. Some sort of incompatibility perhaps?

cyqsimon:
Also notably, on EL8 chroots the failure is a segfault, whereas on other chroots it's a bad exit status.

Not sure what's going on here.

binwiederhier:
I updated the frontend tool chain to newer versions. Your guess is as good as mine

cyqsimon:
I'll check if node-24 works and report back.

Also the [development docs](https://docs.ntfy.sh/develop/#install-dependencies) probably should be updated. I doubt these ancient versions would work.

cyqsimon:
Okay this is getting very strange. node-22 does work, just not in COPR environment somehow.

So I tested two cases:
1. RockyLinux 10 container on ArchLinux LTS kernel with node-22: works.
2. RockyLinux 9 container on RockyLinux 9 kernel with node-22: works.

So now I'm just flustered. This all makes very little sense. I mean vite failing with a segfault is strange enough nevermind this.

I guess I will go to talk to COPR people about this.

binwiederhier:
Thanks for digging into this. Since node-22 works everywhere except inside the COPR chroots (and even segfaults on EL8), this really looks like a COPR/sandbox-environment issue rather than something in ntfy's build itself -- so I agree that's best raised with the COPR folks.

[...]

PR #1896, opened 2026-08-09
Bump rollup

cyqsimon:
I ran into build failures on RHEL8 due to glibc version requirements in the shipped binaries of rollup. This issue was also encountered upstream and has already been fixed, so the only action necessary on our end is to bump the patch version of rollup.

See:

- https://copr.fedorainfracloud.org/coprs/cyqsimon/ntfysh/build/10835139/
- https://github.com/rollup/rollup/releases#release-v4.62.4
- https://github.com/rollup/rollup/pull/6467

This only affects 2.27.0, so if you can put out a patch release that'd be great. But it's up to you.

P.s. #1803 is still a problem, but I've worked around it for the time being by adding a build patch that downgrades to vite-7.

### DIFF
--- a/web/package-lock.json
+++ b/web/package-lock.json
@@ -2198,6 +2198,26 @@
         }
       }
     },
+    "node_modules/@napi-rs/lzma-linux-x64-gnu": {
+      "version": "1.5.1",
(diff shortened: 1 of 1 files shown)


cyqsimon:
Superceded.

PR #1931, opened 2026-09-02
Support RFC 3339 timestamps for delays

ALPHACOM-Brehmer:
Add support for RFC 3339 timestamps in the Delay property, allowing messages to be scheduled at a precise point in time.

### DIFF
--- a/Dockerfile-build
+++ b/Dockerfile-build
@@ -49,6 +49,9 @@ ADD ./attachment ./attachment
 ADD ./mail ./mail
 ADD ./s3 ./s3
 ADD ./action ./action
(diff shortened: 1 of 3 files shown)


ALPHACOM-Brehmer:
I never make commits directly to the main branch, since I believe it should be reserved for releases.
If this conflicts with your personal branching workflow, feel free to rebase the commit or merge it using a fast-forward merge.

Dockerfile-build
ALPHACOM-Brehmer: Why do I have to do this, to make the Docker build and tests working?

